PR #2, opened 2021-02-21
fix: use models the sequelize way

coderbyheart:
So TypeScript types are properly resolved.
See https://github.com/RobinBuschmann/sequelize-typescript\#build-and-create


src/resolvers.ts
```diff
@@ -5,16 +5,14 @@ import Group from './models/group'
 const resolvers: Resolvers = {
   Query: {
     groups: async (parent, args, ctx) => {
-      // @ts-ignore
-      const groupList: Group[] = await sequelize.models.Group.findAll()
+      const groupList: Group[] = await Group.findAll()
```
coderbyheart, line 8: It's unfortunate that this depends on globals (the sequelize instance is magically conjured into existence here), but I remember this to be the *sequelize way* of doing things.

PR #16, opened 2021-02-22
Add integration test for groups API

### LINKED ISSUE #7: Set up servers-side testing framework
Done when: server-side code can be unit tested and integration tested with the test database. Recommend using Jest for a testing framework but open to other options.

Add test runs to CI: https://github.com/distributeaid/shipment-tracker/blob/saga/.github/workflows/ci.yml#L33-L34

bion:
Closes #7 

- [x] passing API integration tests that exercise the database
- [x] tests added to CI
- [x] make SQL logging configurable, disabled by default
- [x] set up pg on CI

### DIFF
--- a/src/sequelize.ts
+++ b/src/sequelize.ts
@@ -9,5 +9,6 @@ export const sequelize = new Sequelize({
   username: config.username,
   password: config.password,
   dialect: config.dialect,
+  logging: process.env.NODE_ENV?.toUpperCase() !== 'TEST',
   models: [__dirname + '/models'],
 })
--- a/src/sequelize.ts
+++ b/src/sequelize.ts
@@ -1,6 +1,6 @@
 import { Sequelize } from 'sequelize-typescript'
 
-const env = process.env.NODE_ENV || 'development'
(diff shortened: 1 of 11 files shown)


src/sequelize.ts
```diff
@@ -1,6 +1,6 @@
 import { Sequelize } from 'sequelize-typescript'
 
-const env = process.env.NODE_ENV || 'development'
+const env = process.env.DB_ENV || process.env.NODE_ENV || 'development'
```
andrewjacobson-outreach, line 6: Haven't seen `DB_ENV` used before. When does this get set instead of `NODE_ENV`?

andrewjacobson-outreach (approved):
LGTM!

src/sequelize.ts
bion: It's just for CI since the DB connection is slightly different for CI vs local test. I'll add a comment to clarify.

andrewjacobson-outreach: Ah got it. Sounds good 👍

PR #17, opened 2021-02-22
docs: do not enforce precise Node.js version

coderbyheart:
In Node.js project there is typically no need to enforce a specific
version of Node.js, especially when targeting an LTS release
because the stability of the LTS release is very good and you
definetly want stability and security updates.


.node-version
```diff
@@ -1 +0,0 @@
-14.15.5
```
bion: Is it possible to also  use `>=14.0.0` here?

dbalatero: I would prefer to revert this delete. Having this file here makes `nodenv` do the Right Thing when you `cd` into the directory. Just leave the README and `package.json` looser, and have some version opinion in `.node-version` and we'll have the best of both worlds.

bion: That sounds reasonable to me

PR #41, opened 2021-02-26
Added some layout components with DA branding

deammer:
### Changes:

- Added some new layout components to wrap the contents of the app
- Added the Distribute Aid logo

### Pixels:
Application layout


frontend/src/components/PageHeader.tsx
```diff
@@ -0,0 +1,25 @@
...
+interface Props {
+  title?: ReactNode
+}
+
+/**
+ * A full-width element with branding and important links.
+ */
+const PageHeader: FunctionComponent<Props> = ({ title }) => {
+  return (
+    <header className="py-2" style={{ background: '#112C71' }}>
```
bion: suggestion (non-blocking): extract the color value to a common constants file?

bion (approved):

frontend/src/components/PageHeader.tsx
deammer: Yeah I went back and forth on that, since we don't haven't fleshed out the color palette. A variable is better though 👍

PR #42, opened 2021-02-26
Server auth

### LINKED ISSUE #6: Integrate backend with auth0 for authentication
Done when: the server only accepts graphql requests from clients authenticated with auth0.

bion:
Walk through video: https://www.loom.com/share/b3e39cc0cf964369bfa8e228f53a7c0c

Authenticates all GraphQL API calls with Auth0.
Devs can opt to skip server-side auth by setting an env var: `SKIP_AUTH=SKIP`

Closes #6


src/sequelize.ts
```diff
@@ -12,6 +12,6 @@ export const sequelize = new Sequelize({
   username: config.username,
   password: config.password,
   dialect: config.dialect,
-  logging: process.env.NODE_ENV?.toUpperCase() !== 'TEST',
+  logging: process.env.NODE_ENV?.toUpperCase() !== 'TEST' && console.log,
```
bion, line 15: 💁🏻 this fixes a deprecation warning from sequelize where the value for `logging` should be either `false` or a function for logging

src/server.ts
```diff
@@ -1,23 +1,18 @@
...
+import apolloServer from './apolloServer'
+
+dotenv.config()
 
 const app = express()
 app.use('*', cors())
 app.use(compression())
 
-server.applyMiddleware({ app, path: '/graphql' })
+apolloServer.applyMiddleware({ app, path: '/graphql' })
```
bion, line 15: 💁🏻 we've got both an internal Apollo "server" abstraction and an Express server that actually binds a network socket and handles HTTP requests, renaming this seemed like it might suggest that a little more clearly

.env.example
```diff
@@ -0,0 +1,3 @@
+# AUTH_MODE
+# skips authenticating GraphQL requests with Auth0 when set to "SKIP"
+AUTH_MODE=SKIP
```
bion, line 3: 💁🏻 I'm open to other thoughts on this but for now skipping auth for GraphQL endpoints by default seemed unobtrusive while get the full permissions story sorted out

andyjac: Seems good to me 👍

src/sequelize.ts
andyjac: Ah! I was wondering how to fix this 😄

src/server.ts
andyjac: Good call 👍

andyjac (approved):
LGTM! Good stuff :)

PR #43, opened 2021-02-28
Add create shipment and list shipments APIs

### LINKED ISSUE #19: Shipment mutate API
Background: see `Shipment` type in proposed [GraphQL schema here](https://www.notion.so/distributeaid/Technical-requirements-c2fd190e0a8d4f708119c6944fa654dd#ee90467d687d4d158a5a97d131c7a862).

Done when graphql API supports creating and updating shipments

andyjac:
Fixes #19 

Screen Shot 2021-02-27 at 4 27 01 PM

### Features added:
- Adds ability to create a shipment
- Adds ability to list shipments

#### Other notable changes:
- Added `graphql-scalars` package, which bring in a bunch of optional useful custom scalar types. I'm using their `Date` scalar for all date values
- Added `sendingHubId` and `receivingHubId` to the GraphQL schema for shipment
- Tried to start moving resolvers to their own separate directories to keep the root resolver tidier
- Tried my hand at creating the first couple custom relationship resolvers for `sendingHub` and `recievingHub` (see: [example](https://github.com/distributeaid/shipment-tracker/pull/43/files#diff-627c4e608ef4924ba733b72e31ec9ae51bc7972a17796af1251d73bb43f6c112R60-R66))

#### Follow-up items
Once we have consensus on the schema proposals I've put into this PR I'll update our documentation in Notion to reflect the changes.

__NOTE:__ I will tackle `listShipment(id: Int)` and `updateShipment(input: ShipmentInput)` work in a second PR since this one is already getting large.


frontend/.gitignore
```diff
@@ -22,29 +22,4 @@ npm-debug.log*
 yarn-debug.log*
 yarn-error.log*
 
-.vscode# See https://help.github.com/articles/ignoring-files/ for more about ignoring files.
```
andyjac, line 25: There were some duplicate lines in the frontend .gitignore (probably just copy-past issue) so I'm just removing them here.

tsconfig.json
```diff
@@ -8,7 +8,8 @@
     "lib": ["esnext"],
     "moduleResolution": "node",
     "strict": true,
-    "target": "ES6"
+    "target": "ES6",
+    "strictPropertyInitialization": false
```
andyjac: Disabling so our models don't need to append `!` to every property. Someone tell me if this is a really bad idea...

.gitignore
```diff
@@ -1,2 +1,27 @@
-node_modules/
-dist/
+# See https://help.github.com/articles/ignoring-files/ for more about ignoring files.
```
andyjac, line 1: Adding @deammer's awesome .gitignore goodness to our base file.

.tool-versions
```diff
@@ -0,0 +1 @@
+nodejs 14.15.5
```
andyjac, line 1: Tools like `asdf` can read this to determine which version to use when `cd`ing into the directory.

schema.graphql
```diff
@@ -1,12 +1,63 @@
...
+  # label{year,month} are for identification purposes only e.g. for deriving
+  # display names like "UK-2021-03".
+  labelYear: Int!
+  labelMonth: Int!
+  offerSubmissionDeadline: Date
+  status: ShipmentStatus!
+  sendingHubId: Int!
+  sendingHub: Group!
+  receivingHubId: Int!
+  receivingHub: Group!
```
andyjac, line 56: Adding properties for the ids so our schema aligns with the db model.

src/models/shipment.ts
```diff
@@ -0,0 +1,58 @@
...
+
+import { ShipmentStatus, ShippingRoute } from '../server-internal-types'
+import Group from './group'
+
+@Table({
+  timestamps: true,
+})
+export default class Shipment extends Model {
+  @Column(DataType.STRING)
+  public shippingRoute: ShippingRoute
```
andyjac: Without the explicit `STRING` type here, sequelize complains becuase `ShippingRoute` is a TS enum. If someone knows a better way around this let me know!

src/models/shipment.ts
```diff
@@ -0,0 +1,58 @@
...
+  public labelYear: number
+
+  @Column
+  public labelMonth: number
+
+  @Column
+  public offerSubmissionDeadline: Date
+
+  @Column(DataType.STRING)
+  public status: ShipmentStatus
```
andyjac: Same enum issue as above.

src/resolvers/shipment/index.ts
```diff
@@ -0,0 +1,68 @@
...
+    !input.sendingHubId ||
+    !input.receivingHubId ||
+    !input.status
+  ) {
+    throw new UserInputError('Shipment arguments invalid', {
+      invalidArgs: Object.keys(input).filter(
+        (key) => !input[key as keyof ShipmentInput],
+      ),
+    })
+  }
```
andyjac, line 40: I wasn't sure the best way to handle validation. Happy to have others input here as this is a pretty ugly method.

schema.graphql
```diff
@@ -1,12 +1,63 @@
+scalar Date
+
 type Group {
   id: Int
   name: String
 }
 
+input GroupInput {
```
andyjac, line 8: Found a lot of examples online using the `input` type to declare query or mutation inputs that way they can be reused. For example, create and update mutations may often have the same params save for update needing an id:

```graphql
addShipment(input: ShipmentInput!): Shipment
updateShipment(id: Int!, input: ShipmentInput!): Shipment
```

frontend/.gitignore
deammer: Wow, my bad, thank you for fixing that!

Could you add the bottom 2 lines back in though?

```
# CSS files output
*.output.css
```

andyjac: Oh whoops! Yeah, let me add those back in.

andyjac: ✅

api-types.ts
```diff
@@ -9,25 +9,77 @@ export type Scalars = {
...
+  Open = 'OPEN',
+  Staging = 'STAGING',
+  InProgress = 'IN_PROGRESS',
+  Complete = 'COMPLETE',
+  Abandoned = 'ABANDONED'
+}
+
+export type Shipment = {
+  __typename?: 'Shipment';
+  id?: Maybe<Scalars['Int']>;
```
coderbyheart: What is the advantage over using:

```typescript
     id?: number
```

coderbyheart: In addition, this semantically means, that there maybe maybe an int, because the the property is marked optional using the TypeScript `?` and even if the value is present, it could not contain an id. Why these duplicate semantics?

api-types.ts
```diff
@@ -9,25 +9,77 @@ export type Scalars = {
...
 };
 
+export type GroupInput = {
+  name?: Maybe<Scalars['String']>;
+};
+
 export type Query = {
   __typename?: 'Query';
   listGroups: Array<Maybe<Group>>;
+  listShipments: Array<Maybe<Shipment>>;
```
coderbyheart: Why return a list of maybe shipments? Here a list of 0 or more Shipments is returned. Why would you return a list of items that can be a shipment or an absent shipment?

schema.graphql
```diff
@@ -1,12 +1,62 @@
...
+  name: String
+}
+
 type Query {
   listGroups: [Group]!
+  listShipments: [Shipment]!
```
coderbyheart:
```suggestion
  listShipments: [Shipment!]!
```

otherwise `[null]` is allowed.

tsconfig.json
coderbyheart: This disables this check for all class properties in the code base, not just for models. What is the issue with appending `!` (which makes this very explicit)?

src/resolvers/shipment/index.ts
coderbyheart: I recommend to use validate using JSON schema, which is language-agnostic and has very powerful validation rules.
[ajv](https://github.com/ajv-validator/ajv) is a good choice for that.
Combining it with [@sinclair/typebox](https://github.com/sinclairzx81/typebox) then turns the JSON schema into TypeScript types.
[Here is an example from a lambda where I use it](https://github.com/NordicSemiconductor/asset-tracker-cloud-aws-js/blob/e686209377b8864356d3147968e5b73a6a332b5b/cellGeolocation/httpApi/addCellGeolocation.ts#L60)

coderbyheart: In addition, this code should also then validate that the passed ids are valid, e.g. the request should be rejected if the receiving hub does not exist, and/or the user does not have sufficient permissions to create a shipment.

deammer: Suggestion:  let's leave permissions out of this PR since we haven't set up Auth0 yet (Bion is working on it in #42).

api-types.ts
andyjac: Yeah, this file is generated by [`graphql-codegen`](https://graphql-code-generator.com/), which is where these `Maybe` types come from. Getting our `schema.graphql` in the shape we want is what will influence the output here. I will mark `id` as non-null with `id: Int!` which should shore some of this up.

- [x] TODO: Make `id` `Int!` in `schema.graphql`

api-types.ts
andyjac: This will be updated when I address your other comment to make `listShipments` return `[Shipment!]!`.

schema.graphql
andyjac: Yup, good call. Will address this.

- [x] TODO: Make `listShipments` return `[Shipment!]!`

tsconfig.json
andyjac: Yeah, fair point. Not a major issue to be explicit in our model files. Will revert this.

- [x] TODO: Revert `"scriptPropertyInitialization": false` change

src/resolvers/shipment/index.ts
andyjac: @coderbyheart good point about validating the existence of the sending and receiving hub. I can add that to this PR. I agree with @deammer that permission validation can probably wait for @bion's work, but I'll also let @bion weigh in on this as well.

This JSON schema validation seems like a good way to go. I'll be doing a second PR to add two more shipment CRUD resolvers and would be happy to dig into better validation then.

- [x] TODO: Validate that sending and receiving hub exist before db write

bion: @coderbyheart's recommendations of ajv with @sinclair/typebox sound good to me for improving validations.

For permissions yes let's wait until authentication is available before implementing authorization. Hopefully my auth0 PR will be done soon!

src/models/shipment.ts
bion: Sounds reasonable to me. I prefer strings for representing enums in DB columns since they're easier to read and decipher what a query is doing or what's in a row when accessing the DB directly.

docs/codegen.md
```diff
@@ -0,0 +1,19 @@
+# Codegen
+
+In order to generate TypeScript typings from our graphql schema we rely on a tool called [graphql-codegen](https://graphql-code-generator.com/).
+
```
bion, line 4:
```suggestion

When you introduce a change to the project's GraphQL schema (the `schema.graphql` file in the root directly) you will need to re-run the codegen tool. Makes sure to check in the generated code changes along with the change to the schema file in your PR.

```

README.md
```diff
@@ -62,6 +62,11 @@ If you run into problems setting up your development environment please create a
 
 See the README in the `frontend` directory for instructions on setting up for front end development.
 
+## Other project docs
+
+- [Database migrations](./docs/migrations.md)
+- [Codegen](./docs/codegen.md)
+
```
bion, line 69: Thank you for improving the documentation!

src/models/group.ts
```diff
@@ -1,26 +1,37 @@
...
 
   @CreatedAt
   @Column
-  public createdAt!: Date;
+  public readonly createdAt!: Date
 
   @UpdatedAt
   @Column
-  public updatedAt!: Date;
+  public readonly updatedAt!: Date
```
bion, line 36: 👍🏻

src/resolvers.ts
```diff
@@ -1,32 +1,34 @@
...
+    addGroup,
+    addShipment,
+  },
 
-      return { id: group.id, name: group.name }
-    },
+  // Custom Resolvers
+  Shipment: {
+    sendingHub,
+    receivingHub,
```
bion, line 31: 👍🏻 definitely way more extensible this way

src/resolvers/group/index.ts
```diff
@@ -0,0 +1,29 @@
+import { UserInputError } from 'apollo-server'
+
+import { QueryResolvers, MutationResolvers } from '../../server-internal-types'
+import { sequelize } from '../../sequelize'
+import Group from '../../models/group'
+
+const groupRepository = sequelize.getRepository(Group)
+
+// Group query resolvers
+const listGroups: QueryResolvers['listGroups'] = async () => {
```
bion, line 10: question: could this be `QueryResolvers.listGroups` instead or will ts not like that

src/resolvers/shipment/index.ts
```diff
@@ -0,0 +1,82 @@
...
+
+  const sendingHub = await groupRepository.findByPk(input.sendingHubId)
+  if (!sendingHub) {
+    throw new ApolloError('Sending hub not found')
+  }
+
+  const receivingHub = await groupRepository.findByPk(input.sendingHubId)
+  if (!receivingHub) {
+    throw new ApolloError('Receiving hub not found')
+  }
```
bion, line 53: suggestion (non-blocking): I think retrieving the sending and receiving hubs could be done concurrently rather than in strict serial order by doing this:

```typescript
  const sendingHubPromise = groupRepository.findByPk(input.sendingHubId)
  const receivingHubPromise = groupRepository.findByPk(input.sendingHubId)
  
  const sendingHub = await sendingHubPromise
  if (!sendingHub) {
    throw new ApolloError('Sending hub not found')
  }
  
  const receivingHub = await receivingHubPromise
  if (!receivingHub) {
    throw new ApolloError('Receiving hub not found')
  } 
```

src/resolvers/shipment/index.ts
```diff
@@ -0,0 +1,82 @@
...
+}
+
+const receivingHub: ShipmentResolvers['receivingHub'] = async (parent) => {
+  const receivingHub = await groupRepository.findByPk(parent.receivingHubId)
+  if (!receivingHub) {
+    throw new ApolloError('Receiving hub not found')
+  }
+
+  return receivingHub
+}
```
bion, line 83: thought: aaaaaaaah so this is how graphql resolves associations. I've been wondering what this was gonna like look. Thanks!

bion (approved):
Code LGTM! If you'd rather merge and file a ticket for bringing in libraries for the validation stuff feel free to do that, up to you.

docs/codegen.md
andyjac: Love it! I'll make this change.

src/resolvers/group/index.ts
andyjac: I don't think we can use dot notation unfortunately. Let me look into this more. I don't particularly like it either, but it was the way I could figure out how to get the typings correct. Good news is even with the string access we still get autocomplete and type validation so entering invalid keys will still break the build. But let me research more here...

src/resolvers/shipment/index.ts
andyjac: Will make this change. Much better.

src/resolvers/shipment/index.ts
andyjac: I was wondering the same thing!

src/resolvers/shipment/index.ts
andyjac: Ticket created to track API input validation work: https://github.com/distributeaid/shipment-tracker/issues/47

PR #44, opened 2021-02-28
Display a list of all the groups

deammer:
### Loom recording ▶️

I'm trying this thing called Loom where I record myself going over PR changes. I'm sorry that I say "um" so much and that my webcam wasn't working, but I hope it's somewhat useful. Feedback is welcome!

https://www.loom.com/share/4474c3d6b46b47c48d632ac4499eaf58

### Changes:

* added a new `/groups` page that lists the groups
* added some fixtures for developing the UI faster
* added [`react-table`](https://react-table.tanstack.com/docs/overview) for easy sorting and filtering. I love this module because it provides functionality without enforcing any rendering or styling pattern. The type defs are terrible though and [require some manual definitions](https://github.com/DefinitelyTyped/DefinitelyTyped/tree/master/types/react-table)  so I might reconsider in the future 🤔 
* added [`classnames`](https://github.com/JedWatson/classnames) 

### Pixels:

Screen Shot 2021-02-27 at 6 17 05 PM


frontend/README.md
```diff
@@ -73,3 +73,12 @@ In order of operation:
 We follow Create React App's [approach to environment variables](https://create-react-app.dev/docs/adding-custom-environment-variables/#adding-development-environment-variables-in-env). There is a `.env` file **checked in** that contains non-sensitive default variables. You can create a `.env.local` file on your own machine and it should **not** be checked in.
 
 Environment variables must be prefixed with `REACT_APP_`.
+
+## Notes
+
+It might be beneficial to move the frontend into its own repo. List of reasons why:
```
jtfairbank: **Question (non-blocking):** Make this a `discussion` issue?

I don't have a specific preference as to 1 or 2 repos.  It's come up before but the project's context has changed now that we're using Node on the backend.  By making this an issue devs working on the project can discuss it in one spot, resolve it once consensus is reached, even re-open it if new trade-offs are discovered down the road.

As decisions are made at this point in the project lifespan, it may also be wise to document them.  @coderbyheart has a nifty "Architectural Decision Record" format we could try out once things begin to stabalize!

jtfairbank (commented):
Code looks ok to me!  I haven't run it yet tho (still need to setup the project), and I'm a bit rusty on current JS best practices.

frontend/src/types/group.ts
```diff
@@ -0,0 +1,20 @@
...
+}
+
+export type Group = {
+  id: number
+  name: string
+  groupType: GroupType
+  primaryLocation: GroupLocation
+  primaryContactName: string
+  website?: string
+}
```
bion: Suggestion (blocking): I think we should be able to import these types directly from the common schema. From the video it sounds like a front end build or dev tool requires that all files for it be available in the `frontend` subdir. That's doable by editing the `codegen.yaml` in the root dir. Try changing like 9 from `./api-types.ts:` to `frontend/api-types.ts:`or `frontend/src/api-types.ts:` (whatever you prefer).

I'd be really nice if we could use a common source of truth between the client and server for defining the shape of data.

frontend/src/pages/groups/GroupList.tsx
```diff
@@ -0,0 +1,114 @@
+import { FunctionComponent, useMemo } from 'react'
+import { useSortBy, useTable } from 'react-table'
+import cx from 'classnames'
```
bion, line 3: nitpick (non-blocking): thoughts on calling this variable `cs` or `classNames` or somesuch? `cx` doesn't evoke "class names" for me.

frontend/README.md
bion: I'm happy to have a discussion on this and agree that we should move that to an issue and perhaps schedule a time to talk about it. fwiw I'm pretty accustomed to working in monorepos and see some benefits that I'm hesitant to let go of.

frontend/src/pages/groups/GroupList.tsx
```diff
@@ -0,0 +1,114 @@
...
+import LayoutWithNav from '../../layouts/LayoutWithNav'
+import groupsFixture from '../../fixtures/groups'
```
bion, line 5: Request (non-blocking): please add an issue for loading this from the server

frontend/src/utils/format.ts
```diff
@@ -0,0 +1,14 @@
...
+export function formatGroupType(type: GroupType) {
+  switch (type) {
+    case GroupType.DAHub:
+      return 'DA hub'
+    case GroupType.ReceivingGroup:
+      return 'Receiving group'
+    case GroupType.SendingGroup:
+      return 'Sending group'
+    default:
+      return type
```
bion: Suggestion (non-blocking): my inclination is that `switch` statements on enums should be exhaustive and throw by default to catch unintended behavior as early as possible. Open to hearing other thinking on it though!

```suggestion
      throw new Error(`Unknown GroupType: ${type}`)
```

frontend/src/pages/groups/GroupList.tsx
```diff
@@ -0,0 +1,114 @@
...
+                      className={cx(
+                        'p-2 first:pl-6 last:pr-6 bg-gray-50 text-gray-700 text-left font-semibold',
+                        {
+                          'hover:bg-gray-100': column.canSort,
+                          'bg-gray-100': column.isSorted,
+                        },
+                      )}
+                      title={column.canSort ? 'Sort rows' : ''}
+                    >
+                      {column.render('Header')}
```
bion, line 71: thought: ah, I see how type checking could be tricky with this library if the API involves passing strings keys in like this

frontend/src/types/react-table-config.d.ts
```diff
@@ -0,0 +1,49 @@
...
+
+  export interface Cell<
+    D extends Record<string, unknown> = Record<string, unknown>,
+    V = any
+  > {}
+
+  export interface Row<
+    D extends Record<string, unknown> = Record<string, unknown>
+  > {}
+}
```
bion, line 49: Question: do you have a sense of how much updating this file will need?

frontend/src/types/group.ts
deammer: I agree about the common source of truth, I'll give it a shot 👍

README.md
```diff
@@ -18,7 +18,7 @@ First, please read our [code of conduct](https://www.notion.so/distributeaid/Cod
 
 ### Issue tracking
 
-Development tasks are managed in the github issues for this repository. The issues themselves are fairly light on detail in favor of a simple description of scope (i.e. the conditions for the task being considered "done"). For specifics on task requirements, please reference the [project requirements document](https://www.notion.so/distributeaid/Technical-requirements-c2fd190e0a8d4f708119c6944fa654dd)  early and often.
+Development tasks are managed in the github issues for this repository. The issues themselves are fairly light on detail in favor of a simple description of scope (i.e. the conditions for the task being considered "done"). For specifics on task requirements, please reference the [project requirements document](https://www.notion.so/distributeaid/Technical-requirements-c2fd190e0a8d4f708119c6944fa654dd) early and often.
```
deammer: Thanks, Prettier

frontend/src/types/group.ts
deammer: @bion I made the changes you suggested in https://github.com/distributeaid/shipment-tracker/pull/44/commits/9fffcb9b13a9141f61e9fe3ce2b0050dd9496e74.

frontend/src/pages/groups/GroupList.tsx
deammer: There's [a bit of history there](https://github.com/JedWatson/classnames/issues/157) 🤓 I am lazy and find `classNames` to be too long to type and a little difficult to distinguish from `className` at a glance (bad eyesight over here.) I wouldn't mind using `cn` or `cs`, but I've seen `cx` used in enough codebases that I think it's an established convention. @andyjac what do you think?

frontend/src/types/react-table-config.d.ts
deammer: We would need to update these typedefs if we used more of the available plugins. Right now, we only use `useSortBy` (see code below), but I could see us add [`useFilters`](https://react-table.tanstack.com/docs/api/useFilters), for example.

```ts
const {
  getTableProps,
  getTableBodyProps,
  headerGroups,
  rows,
  prepareRow,
} = useTable({ columns, data }, useSortBy) // <- last argument here
```

deammer: I'm hoping they release v8 of the package soon because it's written entirely in TypeScript. Even then though, I feel like the benefits outweigh the drawbacks.

frontend/README.md
deammer: I will give it a little more thought and move it to an issue if needed 👍

README.md
```diff
@@ -62,6 +62,16 @@ If you run into problems setting up your development environment please create a
...
+### Type definitions
+
+We want to maintain a single source of truth for our type definitions. We use `graphql-codegen` for that purpose.
+
+The command below will output TypeScript definitions that can be used by the backend and frontend.
+
+```shell
+yarn codegen
+```
+
```
bion, line 83: 👍🏻

frontend/src/fixtures/groups.ts
```diff
@@ -0,0 +1,96 @@
...
+type GroupPlaceholder = Group & {
+  groupType: GroupType
+  primaryLocation: {
+    countryCode?: string
+    townCity: string
+  }
+  primaryContact: {
+    name: string
+  }
+}
```
bion, line 14: 👍🏻

frontend/src/pages/groups/GroupList.tsx
deammer: 👍  https://github.com/distributeaid/shipment-tracker/issues/46

frontend/src/pages/groups/GroupList.tsx
bion: Ah ok, sounds like I'm the odd one out then 😆, I can get used to `cx` if it's conventional

frontend/src/types/react-table-config.d.ts
bion: Thanks for explaining, helpful to know!

frontend/src/fixtures/groups.ts
```diff
@@ -1,8 +1,13 @@
...
+import { Group } from '../types/api-types'
 
 // TODO remove the placeholder fields when the issue below is resolved
 // https://github.com/distributeaid/shipment-tracker/issues/45
+export enum GroupType {
+  DAHub = 'DA_HUB',
+  ReceivingGroup = 'RECEIVING_GROUP',
+  SendingGroup = 'SENDING_GROUP',
+}
+
```
bion, line 5: Suggestion (blocking): move the `GroupType` enum into the project's `schema.graphql` file

bion (approved):

codegen.yaml
```diff
@@ -9,3 +9,14 @@ generates:
...
+  # Generate the types used by the frontend, because react-scripts requires them
+  # to exist in the src directory.
+  ./frontend/src/types/api-types.ts:
+    plugins:
+      - add:
+          content: "/**\n
+            * ⚠️ These type definitions are generated automatically by graphql-codegen.\n
+            * The generation is configurated in codegen.yaml. Visit the docs for more information:\n
+            * https://graphql-code-generator.com/docs/getting-started/index\n
+            */\n"
```
andyjac, line 21: Thank you for adding this!! 🙏

frontend/src/pages/groups/GroupList.tsx
andyjac: Yeah, I've seen `cx` in other projects also, so I think this is pretty standard and would be friendly to other front-end folks as well.

andyjac (approved):

frontend/src/fixtures/groups.ts
deammer: Done!

Issue #9, opened 2021-02-21
Group mutate API


bion:
Background: see `Group` type in proposed [GraphQL schema here](https://www.notion.so/distributeaid/Technical-requirements-c2fd190e0a8d4f708119c6944fa654dd#ee90467d687d4d158a5a97d131c7a862).

Done when graphql API supports creating and updating groups

Sgtpluck:
@bion it's unclear from the notation -- is Location and ContactInfo also objects in the database or JSON? I'm assuming they are their own tables, which means they have to be created before Group can be modified, yes?

bion:
@Sgtpluck seems like it could be implemented either way. An additional table seems like a reasonable approach. I'd be fine with either so long as the implementation supports the GraphQL-defined schema.

bion:
@Sgtpluck thinking about it a little more I wonder if storing those in JSON columns wouldn't be the easiest. I can't think of a strong reason to prefer a normalized DB schema.

deammer:
@Sgtpluck I just merged #44 where I added the `groupType` field to the `Group` model, so I recommend pulling the latest code when you work on this.

Sgtpluck:
closed with #51

PR #51, opened 2021-03-03
group mutate

Sgtpluck:
Closes issue #9.

Screen Shot 2021-03-06 at 9 38 16 AM
Features added:

- Updates Groups to have attributes as defined [in proposed GraphQL](https://www.notion.so/distributeaid/Technical-requirements-c2fd190e0a8d4f708119c6944fa654dd#ee90467d687d4d158a5a97d131c7a862).

Other changes:
- Updated tests to make it easier to update group attributes.

### DIFF
--- a/db/migrations/20210302222654-update-groups.js
+++ b/db/migrations/20210302222654-update-groups.js
@@ -1,36 +1,39 @@
-'use strict';
+'use strict'
 
 module.exports = {
   up: async (queryInterface, Sequelize) => {
-    // TODO: make this a transaction
-    await Promise.all([
-      queryInterface.addColumn('Groups', 'primaryLocation',
-        {
+    return queryInterface.sequelize.transaction((t) => {
+      return Promise.all([
+        queryInterface.addColumn('Groups', 'primaryLocation', {
           allowNull: false,
           type: Sequelize.JSONB,
-        }
-      ),
-      queryInterface.addColumn('Groups', 'website', { type: Sequelize.STRING }),
-      queryInterface.addColumn('Groups', 'primaryContact',
-        {
+        }),
+        queryInterface.addColumn('Groups', 'website', {
+          type: Sequelize.STRING,
+        }),
+        queryInterface.addColumn('Groups', 'primaryContact', {
           allowNull: false,
           type: Sequelize.JSONB,
-        }
-      ),
-      queryInterface.changeColumn('Groups', 'name',
-        { allowNull: false, type: Sequelize.STRING }
-      )
-    ]);
+        }),
+        queryInterface.changeColumn('Groups', 'name', {
+          allowNull: false,
+          type: Sequelize.STRING,
+        }),
+      ])
+    })
   },
 
   down: async (queryInterface, Sequelize) => {
-    await Promise.all([
-      queryInterface.removeColumn('Groups', 'primaryLocation'),
-      queryInterface.removeColumn('Groups', 'primaryContact'),
-      queryInterface.removeColumn('Groups', 'website'),
-      queryInterface.changeColumn('Groups', 'name',
-        { allowNull: true, type: Sequelize.STRING  }
-      )
-    ])
-  }
-};
+    return queryInterface.sequelize.transaction((t) => {
+      return Promise.all([
+        queryInterface.removeColumn('Groups', 'primaryLocation'),
+        queryInterface.removeColumn('Groups', 'primaryContact'),
+        queryInterface.removeColumn('Groups', 'website'),
+        queryInterface.changeColumn('Groups', 'name', {
+          allowNull: true,
+          type: Sequelize.STRING,
+        }),
+      ])
+    })
+  },
+}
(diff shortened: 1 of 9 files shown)


andyjac:
Wow, okay so this is not very well documented and I had to shuffle through a dozen or so stack overflow threads and blog posts to find an answer, but apparently `input` types in graphql can only contain scalar values or other input types. This means that I guess you have to duplicate some of these definitions, which seems dumb?

I was able to get things building with the following changes to the `schema.graphql` though:

```graphql
type Location {
  countryCode: String
  townCity: String!
  openLocationCode: String
}

type ContactInfo {
  name: String!
  email: String
  phone: String
  signal: String
  whatsApp: String
}

input LocationInput {
  countryCode: String
  townCity: String!
  openLocationCode: String
}

input ContactInfoInput {
  name: String!
  email: String
  phone: String
  signal: String
  whatsApp: String
}

type Group {
  id: Int!
  name: String!
  groupType: GroupType!
  primaryLocation: Location!
  primaryContact: ContactInfo!
  website: String

  createdAt: Date!
  updatedAt: Date!
}

input GroupInput {
  name: String!
  groupType: GroupType!
  primaryLocation: LocationInput!     # Changed: Location! -> LocationInput!
  primaryContact: ContactInfoInput!   # Changed: ContactInfo -> ContactInfoInput!
  website: String
}
```

andyjac:
@Sgtpluck I've got these changes locally and would happily just push them to your branch. Everything else looks great!

src/resolvers/group/index.ts
```diff
@@ -30,11 +39,10 @@ const addGroup: MutationResolvers['addGroup'] = async (
     groupType: input.groupType,
     primaryLocation: input.primaryLocation,
     primaryContact: input.primaryContact,
-    website: input.website
+    website: input.website || undefined,
```
andyjac: My linter was complaining about `input.website` potentially being `string | null | undefined` when only `string | undefined` were allowed, which seems odd because it's nullable column per the migration file, but maybe we have to call that out explicitly in the model like:

```typescript
@Column
public website?: string | null
```

Not really sure. Still figuring out how these sequelize models work.

schema.graphql
```diff
@@ -22,15 +36,15 @@ type Group {
...
+  updatedAt: Date!
 }
 
 input GroupInput {
   name: String!
   groupType: GroupType!
-  primaryLocation: Location!
-  primaryContact: ContactInfo!
+  primaryLocation: LocationInput!
+  primaryContact: ContactInfoInput!
```
andyjac: This is ultimately the change that got things working 🤷‍♂️

src/resolvers/group/index.ts
Sgtpluck: i'm not sure -- i wasn't having issues with that particularly bit! however i am curious about this type:
` extends Optional<GroupAttributes, 'id'> {}` -- most of our fields are not optional, and it seems like (i am in a rush so maybe reading this incorrectly) this picks the key out of the second argument and makes it non-optional, which is probably not ... a thing we want?

db/migrations/20210302222654-update-groups.js
```diff
@@ -0,0 +1,39 @@
...
+        }),
+        queryInterface.addColumn('Groups', 'website', {
+          type: Sequelize.STRING,
+        }),
+        queryInterface.addColumn('Groups', 'primaryContact', {
+          allowNull: false,
+          type: Sequelize.JSONB,
+        }),
+        queryInterface.changeColumn('Groups', 'name', {
+          allowNull: false,
```
bion, line 19: 👍🏻 good catch

bion (approved):

src/models/group.ts
```diff
@@ -7,16 +7,19 @@ import {
...
-import { GroupType } from '../server-internal-types'
+import { GroupType, Location, ContactInfo } from '../server-internal-types'
 
 export interface GroupAttributes {
   id: number
   name: string
   groupType: GroupType
+  primaryLocation: Location
+  primaryContact: ContactInfo
+  website?: string | null
```
deammer, line 18: Suggestion (non-blocking): this translates to `undefined | string | null`. I think `website?: string` should be enough?

deammer (approved):

src/models/group.ts
Sgtpluck: @deammer it's not enough -- when i had it just as `website?: string` then the linter *sometimes* failed with
`Type 'null' is not assignable to type 'string | undefined'`. 

the only two solutions i found was either this ^^ or pass `website: input.website || undefined` into the create function, and i liked this better. 🤷‍♀️ 

i tried to dig into it, and didn't find a good explanation beyond "graphql scalar types and typescript types sometimes collide because the former is nullable by default and the latter are non-nullable by default"

Sgtpluck: (i would love to find a better solution, that being said. this kinda sucks.)

deammer: Thanks for looking into it!

bion: Thanks for looking into it and the explanation, it seems there are still some rough edges getting sequelize and typescript to cooperate together. I agree that this is better than `website: input.website || undefined` in the create function

PR #50, opened 2021-03-02
Implement client-side auth

deammer:
### Changes
- added `@auth0/auth0-react`
- added the ability to log in and log out using Auth0
- added a public home page
- added a `DropdownMenu` component
- added a bearer token to API calls

### Pixels

|Public home page on mobile|Menu at the top|
|---|---|
|Screen Shot 2021-03-01 at 10 01 43 PM|Screen Shot 2021-03-02 at 9 39 57 PM|

Resolves #48

PR #54, opened 2021-03-03
Create user accounts on authentication, load in graphql server context

### LINKED ISSUE #53: Maintain a UserAccount record for all users who sign in
When someone signs into shipment tracker with Auth0 for the first time, the server should create a `UserAccount` record for them in the database. These records will be used for permissions associations (e.g. group captains).

Done when: logging in for the first time creates a user account record and the apollo server context includes that record on every request.

bion:
Branched off of #50 so the diff will be quite a bit smaller once that's merged.

Closes #53


frontend/src/pages/ApolloDemo.tsx
```diff
@@ -22,7 +23,9 @@ const ApolloDemoPage = () => {
    *    - error: a ApolloError object with some details, in case things fail
    *    - data: the data fetched by your query
    */
-  const { data, loading, error } = useQuery(GROUPS_QUERY)
+  const { data, loading, error } = useQuery(GROUPS_QUERY, {
+    pollInterval: 5000,
+  })
```
bion, line 29: I noticed this kept going after navigating away from the page, any idea how to not do that?

src/findOrCreateProfile.ts
```diff
@@ -0,0 +1,29 @@
...
+    return
+  }
+
+  console.info('creating account for profile', auth.claims.sub)
+
+  const userAccount = await userAccountRepository.create({
+    auth0Id: auth.claims.sub,
+  })
+
+  console.info('created user account', userAccount.id)
```
bion, line 24: leaving info logs in intentionally, we can turn the log level to warn in prod

frontend/src/pages/ApolloDemo.tsx
deammer: The `useQuery` hook returns a `stopPolling` method that should be called when the component unmounts. I'd do it like this:

```jsx
const { data, loading, error, stopPolling } = useQuery(...)

useEffect(function cleanupOnUnmount() {
  // When the component unmounts, the return value of this method will be called
  return stopPolling
}, [])
```

frontend/src/AppRoot.tsx
```diff
@@ -1,28 +1,54 @@
...
 import HomePage from './pages/Home'
 import PublicHomePage from './pages/PublicHome'
 import GroupList from './pages/groups/GroupList'
 import PrivateRoute from './components/PrivateRoute'
 
+const fetchProfile = (token: string) => {
+  return fetch('/profile', {
+    headers: { Authorization: `Bearer ${token}` },
+  })
+}
```
deammer, line 16: Thought: I wish we didn't have to pass the auth token manually here. I would also prefer to use GraphQL for all the requests on the frontend instead of writing custom endpoints on the server. I'll mess with this to see if that's possible.

bion: Part of the reason I made it separate from GraphQL was to keep things simple in the server-side GraphQL context provider -- this way we can assert that there's a user account for all GraphQL requests or throw.

frontend/src/components/ApolloAuthProvider.tsx
```diff
@@ -0,0 +1,54 @@
...
+
+const clientUrl = process.env.REACT_APP_CLIENT_URL
+
+// https://www.apollographql.com/docs/link/links/http/
+const httpLink = new HttpLink({
+  uri: process.env.REACT_APP_SERVER_URL,
+  credentials: 'include',
+})
+
+const ApolloAuthProvider: FunctionComponent = ({ children }) => {
```
deammer, line 22: Refactored the Apollo client instance into a wrapper component that's then used in `AppRoot`.

src/server.ts
```diff
@@ -4,10 +4,10 @@ import { createServer } from 'http'
 import compression from 'compression'
 import cors from 'cors'
 
-import findOrCreateProfile from './findOrCreateProfile'
-
+// Load the env vars before initializing code that depends on them
 dotenv.config()
 
+import findOrCreateProfile from './findOrCreateProfile'
```
deammer, line 10: The order matters here if we wanna skip auth

frontend/src/AppRoot.tsx
deammer: Gotcha. I don't know if this is possible with GraphQL, but it'd be sweet if we could intercept the query below and run some custom code on the server:

```js
const { data } = useQuery(gql`
  query GetCurrentUser {
    user($userId: string) {
      id
      name
    }
  }
`

Comment (deammer) [reply frontend/src/AppRoot.tsx:16]: I feel like it's one of those things we'll iterate on as we learn about the tech though.

File: src/server.ts (hunk 5 above)
Comment (bion) [reply src/server.ts:10]: 🤦🏼‍♂️

File: src/models/user_account.ts [hunk 6]
```diff
@@ -0,0 +1,37 @@
+import {
```
deammer, line 1: Question: do we also need to update the GraphQL schema?

deammer (approved):

src/models/user_account.ts
bion: Not yet, since we're not sending the user account over the wire since for the time being it doesn't contain anything the client doesn't already get directly from auth0

PR #58, opened 2021-03-05
Display a loading screen when checking credentials

### LINKED ISSUE #52: Make the UI look good while loading Auth0
When the Auth0 credentials are being loaded on the client, the UI just says "Loading..." on a blank page. Make that look better!

See #50

deammer:
Resolves #52 

### Changes

- display a loading screen when loading the user's credentials
- removed the "proxy" setting in package.json because it was breaking asset delivery (I don't think it was necessary in the first place)
- some styling improvements

### Pixels

Screen Shot 2021-03-04 at 10 13 58 PM

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -51,6 +51,5 @@
       "last 1 firefox version",
       "last 1 safari version"
     ]
-  },
-  "proxy": "http://localhost:3000"
+  }
 }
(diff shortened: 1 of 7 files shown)


frontend/package.json
```diff
@@ -51,6 +51,5 @@
       "last 1 firefox version",
       "last 1 safari version"
     ]
-  },
-  "proxy": "http://localhost:3000"
```
deammer, line 55: I don't think this was necessary because we explicitly give Apollo the full URL of the endpoint. This would be helpful if we called our API from the same host, like `http://localhost:8080/api/do/something`.

PR #59, opened 2021-03-05
Prevent non-admins from creating shipments

bion:
The behavior change is pretty small here: a single auth check on the only API endpoints that's restricted to the admin so far.

The testing side of it is a bit more involved though. I couldn't find a way to fake auth or the calling user's creds that didn't involve instantiating a completely different apollo server instance per-user. I tried to make that as painless as possible to work with, but it's still a little weird. Very much open to other approaches if someone has a different idea.

### DIFF
--- a/src/tests/shipments_api.test.ts
+++ b/src/tests/shipments_api.test.ts
@@ -2,7 +2,7 @@ import gql from 'graphql-tag'
 import { ApolloServerTestClient } from 'apollo-server-testing'
 
 import Shipment from '../models/shipment'
-import makeTestServer from '../testServer'
+import { makeTestServer, makeAdminTestServer } from '../testServer'
 import { sequelize } from '../sequelize'
 import {
   GroupType,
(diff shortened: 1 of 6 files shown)


src/tests/shipments_api.test.ts
```diff
@@ -70,6 +76,36 @@ describe('Shipments API', () => {
         },
       })
 
+      expect(res.errors).not.toBeUndefined()
+      expect(res.errors).not.toBeEmpty()
+
+      if (res.errors == null || res.errors.length === 0) {
+        return
+      }
```
bion, line 84: This is annoying, I wish the assertions on lines 79 and 80 meant this wasn't necessary but TypeScript didn't agree.

src/tests/shipments_api.test.ts
```diff
@@ -70,6 +76,36 @@ describe('Shipments API', () => {
...
+
+      expect(res.errors[0].message).toEqual(
+        'addShipment forbidden to non-admin users',
+      )
```
andyjac, line 88: Can you use optional chaining here? Might allow you to get rid of the check on line 82:

```typescript
expect(res?.errors?.[0]?.message).toEqual(
  'addShipment forbidden to non-admin users',
)
```

andyjac (approved):
LGTM!

src/tests/shipments_api.test.ts
bion: It does! Thank you!

PR #63, opened 2021-03-07
Deploy to Heroku staging

bion:
Visible here: https://shipment-tracker-dev.herokuapp.com/

- [x] can sign in with auth0
- [x] migrations run automatically on each deploy


package.json
```diff
@@ -6,24 +6,25 @@
   "scripts": {
     "predev": "yarn codegen",
     "dev": "ts-node-dev --no-notify --respawn --transpile-only src/server",
-    "start": "node dist/server",
+    "start": "./node_modules/.bin/nodemon --optimize_for_size --max_old_space_size=920 --gc_interval=100 dist/server.js",
```
bion, line 9: Nodemon to restart the thing if it crashes. Also node flags to tell V8 to GC early and often instead of acting like Chrome and consuming like 1.5 GB before cleaning up after itself.

package.json
```diff
@@ -6,24 +6,25 @@
...
     "clean": "rm -rf dist",
-    "build": "yarn -s run clean && yarn -s run codegen && tsc",
+    "build": "yarn -s run clean && tsc",
```
bion, line 11: We check generated code in so no need to run codegen as a build step

package.json
```diff
@@ -6,24 +6,25 @@
...
     "codegen": "graphql-codegen",
     "codegen:watch": "graphql-codegen --watch",
+    "heroku-postbuild": "yarn build && npx sequelize-cli db:migrate && cd frontend && yarn install && yarn build:styles && yarn build",
```
bion, line 14: `yarn heroku-postbuild` is what Heroku runs instead of `yarn build`

src/server.ts
```diff
@@ -5,7 +5,9 @@ import compression from 'compression'
 import cors from 'cors'
 
 // Load the env vars before initializing code that depends on them
-dotenv.config()
+if (process.env.NODE_ENV !== 'production') {
+  dotenv.config()
+}
```
bion, line 11: We don't use a `.env` file in prod environments

src/server.ts
```diff
@@ -14,6 +16,8 @@ const app = express()
 
 app.get('/profile', findOrCreateProfile)
 
+app.use(express.static('frontend/build'))
```
deammer: Issue: this won't work on nested routes, like https://shipment-tracker-dev.herokuapp.com/groups for example. I think we need to tell Express to render every route statically?

```ts
// I'm not familiar with the `.static` method but this might work?
app.use('*', express.static('frontend/build'))

// Alternate solution:
app.get('*', (req, res) => {
  res.sendFile(path.join(__dirname, 'frontend/build')
}
```

bion: I'll give it a shot

deammer (approved):
Good to go once the routing is fixed 👍

package.json
deammer: Suggestion (non-blocking): can we seed the database as well?

bion: That could get fairly annoying since it would either duplicate whatever data is already seeded from last time or wipe out anything we've created since for testing purposes which would be pretty annoying if someone (especially a non-technical contributor) were testing something on staging and an auto-deploy wiped their data.

src/server.ts
bion: Done

package.json
deammer: You're right 👍

PR #64, opened 2021-03-07
List groups from db

### LINKED ISSUE #27: Admin UI for listing groups
Done when: admins can see a list of all groups

### LINKED ISSUE #46: Load groups from the server
In #44, we added a list of groups that pulls from a fixture. Once we're able to query the groups endpoint with #18, we should do that instead.

deammer:
Resolves #46 and resolves #27.

### Changes

- made the UI list the groups from the database instead of relying on fixtures
- added some fields to the groups seeder
- ⚠️ renamed a seeder file that had a typo -> this will probably cause some weirdness so I'll warn the devs
- added the stub of an admin page for easier navigation

### Pixels

Screen Shot 2021-03-07 at 10 58 31 AM


frontend/src/pages/AdminPage.tsx
```diff
@@ -0,0 +1,19 @@
...
+const AdminPage: FunctionComponent = () => {
+  return (
+    <LayoutWithNav>
+      <div className="text-da-navy-100 flex flex-col space-y-2">
+        <Link to="/groups">Groups</Link>
+        <Link to="/users">Users (in construction)</Link>
+      </div>
+    </LayoutWithNav>
+  )
+}
```
deammer, line 17: Nothing fancy for now, it's just a way to navigate around the site with ease. Will iterate in the future!

Screen Shot 2021-03-07 at 10 59 44 AM

PR #65, opened 2021-03-08
Create groups through the UI

deammer:
### Changes

I set up a lot of new components, sorry about the lengthy list of changes!

- added form components for selects and input fields
- added a `Button` component because it was about time
- added the `react-hook-form` library to make form data easier to handle
- added a page where users can create a new group
- clicking on a group in the `GroupList` page will send users to its view/edit page

**Note:** we don't support editing groups yet, so this PR has some placeholder code

### My philosophy around form components in React

I prefer to create small components that I can then compose into more intricate patterns. That's the approach I'm following with the form components I added with this PR. For example, I added:

- a `<TextInput>` component, which is very close to an `<input type="text">` element
- a `<Label>` component, which is very close to a `<label>` element
- a `<TextField>` component, which is a composition of a `<Label>` above a `<TextInput>`

This approach provides the following advantages:

- lower-level elements work just like you'd expect their HTML equivalent to work, except they're already styled for you
- it's flexible: if I want to display a label on its own, I can
- wrapper elements can abstract some functionality. For example, the `<TextField>` component generates matching `id` and `htmlFor` attributes to its `<TextInput>` and `<Label>`, respectively.

### Screenshot

Screen Shot 2021-03-08 at 9 52 19 PM


frontend/src/pages/groups/GroupForm.tsx
```diff
@@ -0,0 +1,126 @@
...
+   * The callback triggered when the user submits the form
+   */
+  onSubmit: (input: GroupInput) => void
+}
+
+/**
+ * This component encapsulates a form for creating and editing groups.
+ */
+const GroupForm: FunctionComponent<Props> = (props) => {
+  const { register, handleSubmit, reset } = useForm()
```
deammer, line 32: I'm extremely happy with this library. It saved me a ton of manual work while letting me handle the styling however I want.

frontend/src/pages/groups/GroupEditPage.tsx
```diff
@@ -0,0 +1,94 @@
...
+  const { data: originalGroupData, loading: queryIsLoading } = useQuery<{
+    group: Group
+  }>(GET_GROUP, {
+    variables: { id: parseInt(groupId, 10) },
+  })
+
+  // Set up the mutation to update the group
+  // const [updateGroup, { loading: mutationIsLoading }] = useMutation(
+  //   UPDATE_GROUP,
+  // )
```
deammer, line 60: This is where I need access to an `updateGroup` mutation (#67).

frontend/src/components/ButtonIcon.tsx
```diff
@@ -0,0 +1,43 @@
...
+  Children,
+  cloneElement,
+} from 'react'
+import cx from 'classnames'
+
+interface ExplicitProps {
+  /**
+   * Specify whether the icon is before or after the text
+   */
+  placement?: 'before' | 'after'
```
bion: Question: since `placement` is optional, what is the behavior it isn't present?

frontend/src/components/forms/SelectField.tsx
```diff
@@ -0,0 +1,99 @@
...
+
+const SelectField: FunctionComponent<Props> = ({
+  id,
+  label,
+  error = null,
+  options = [],
+  ...otherProps
+}) => {
+  // Create a unique ID in case the use doesn't provide one
+  const [uniqueId] = useState(nanoid())
```
bion, line 74: Thought (non-blocking): it's still weird to me that hooks force this thing where you must call them regardless of whether or not you'll use them

frontend/src/components/forms/SelectField.tsx
```diff
@@ -0,0 +1,99 @@
...
+
+  const fieldId = id || uniqueId
+
+  return (
+    <div className="w-full">
+      <Label htmlFor={fieldId} required={otherProps.required}>
+        {label}
+      </Label>
+      {error && <InlineError>{error}</InlineError>}
+      <SelectInput {...otherProps} hasError={!!error} id={fieldId}>
```
bion, line 86: Suggestion (non-blocking): newlines after L82 and L83 to make this slightly easier to scan

frontend/src/components/forms/SelectInput.tsx
```diff
@@ -0,0 +1,52 @@
...
+      'bg-gray-100': readOnly || disabled,
+      'cursor-not-allowed': disabled,
+    },
+  )
+
+  return (
+    <select
+      {...otherProps}
+      style={{
+        backgroundImage: `url("data:image/svg+xml;charset=utf-8,%3Csvg xmlns='http://www.w3.org/2000/svg' viewBox='0 0 20 20' fill='none'%3E%3Cpath d='M7 7l3-3 3 3m0 6l-3 3-3-3' stroke='%239fa6b2' stroke-width='1.5' stroke-linecap='round' stroke-linejoin='round'/%3E%3C/svg%3E")`,
```
bion, line 39: stringception

frontend/src/pages/groups/GroupCreatePage.tsx
```diff
@@ -0,0 +1,66 @@
+import { FunctionComponent } from 'react'
+import LayoutWithNav from '../../layouts/LayoutWithNav'
+import { Group, GroupInput } from '../../types/api-types'
+import { gql, useMutation } from '@apollo/client'
+import { useHistory } from 'react-router-dom'
```
bion, line 3: nit: here and in a couple other spots it'd be nice to keep the library imports sorted above the relative imports

frontend/src/pages/groups/GroupCreatePage.tsx
```diff
@@ -0,0 +1,66 @@
...
+import GroupForm from './GroupForm'
+
+const ADD_GROUP = gql`
+  mutation Groups($input: GroupInput!) {
+    addGroup(input: $input) {
+      id
+    }
+  }
```
bion, line 13: Suggestion (non-blocking): it'd be nice to pull these into their own files e.g. https://github.com/distributeaid/toolbox/blob/saga/react_ui/src/pages/createChapter.graphql

frontend/src/pages/groups/GroupEditPage.tsx
```diff
@@ -0,0 +1,94 @@
...
+      openLocationCode
+    }
+    primaryContact {
+      name
+      email
+      whatsApp
+      phone
+      signal
+    }
+  }
```
bion, line 25: Yeah especially as they get longer having them in their own files seems preferable.

frontend/src/pages/groups/GroupCreatePage.tsx
```diff
@@ -0,0 +1,66 @@
...
+
+  const [
+    addGroup,
+    { loading: mutationIsLoading, error: mutationError },
+  ] = useMutation<{ addGroup: Group }>(ADD_GROUP)
+
+  const onSubmit = (input: GroupInput) => {
+    // Create the group and then redirect to its view/edit page
+    addGroup({ variables: { input } })
+      .then(({ data }) => {
```
bion, line 27: Question: is `data` type-checked or is it an `any`?

frontend/src/pages/groups/GroupCreatePage.tsx
```diff
@@ -0,0 +1,66 @@
...
+        if (data) {
+          const newGroupId = data.addGroup.id
+          history.push(`/group/${newGroupId}`)
+        }
+      })
+      .catch((error) => {
+        console.log(error)
```
bion: Suggestion (blocking): `console.error` for the error

frontend/src/pages/groups/GroupCreatePage.tsx
```diff
@@ -0,0 +1,66 @@
...
+    { loading: mutationIsLoading, error: mutationError },
+  ] = useMutation<{ addGroup: Group }>(ADD_GROUP)
+
+  const onSubmit = (input: GroupInput) => {
+    // Create the group and then redirect to its view/edit page
+    addGroup({ variables: { input } })
+      .then(({ data }) => {
+        if (data) {
+          const newGroupId = data.addGroup.id
+          history.push(`/group/${newGroupId}`)
```
bion, line 30: Thought (non-blocking): I think it's outside the scope of this PR, but it'd be nice to define routes somewhere common where the server and client could both read them so the server didn't have to guess/assume at what was a valid client-side route.

bion: Defining them that way would also eliminate the class of bugs where the routes in the react router mismatch with hand-typed routes like this one.

frontend/src/pages/groups/GroupCreatePage.tsx
bion: Question: what's the difference between this catch and the `mutationError` variable?

frontend/src/pages/groups/GroupForm.tsx
bion: Yeah this looks pretty slick!

bion (approved):

frontend/src/components/ButtonIcon.tsx
deammer: The icon wouldn't be sized or positioned properly, which isn't good. I made the prop required!

frontend/src/components/forms/SelectField.tsx
deammer: 🤔 we are using the hook here though. It's storing the `uniqueId` variable so that it doesn't change between every render.

frontend/src/pages/groups/GroupCreatePage.tsx
deammer: Yeah because I specified the type of the mutation on line 22:
```ts
const [
    addGroup,
    { loading: mutationIsLoading, error: mutationError },
  ] = useMutation<{ addGroup: Group }>(ADD_GROUP)
                 ^^^^^^^^^^^^^^^^^^^^^
```

frontend/src/components/forms/SelectField.tsx
bion: Not if `id` is present though, yeah?

frontend/src/pages/groups/GroupCreatePage.tsx
deammer: The `error` from the catch block has the type `any`, and it's a different shape from the `mutationError`, which is fully-typed.

I'm using both error variables for different purposes: the `catch` block is used to print the error to the console, while the `mutationError` is displayed to the user.

frontend/src/pages/groups/GroupCreatePage.tsx
deammer: I agree! I've opened #68 to take care of that.

frontend/src/pages/groups/GroupCreatePage.tsx
deammer: I like that a lot, thank you for the suggestion. I've opened #69 to keep track of this.

frontend/src/components/forms/SelectField.tsx
deammer: Right, if an `id` prop is explicitly passed, the `fieldId` variable below will use that instead. But either way, the `uniqueId` variable is instantiated once the component is mounted, and never changes after that.

frontend/src/pages/groups/GroupCreatePage.tsx
deammer: I wonder if there's an editor plugin that can do this automagically 🤔 **Edit:** I see a few ways to do that on the interwebz!

PR #66, opened 2021-03-08
Default new group's captain to creator

bion:
Since non-admins can create groups, we associate newly created groups with their creator as the group's captain.

Also made the profile endpoint return the user account ID and whether or not they're an admin. Over time we'll probably put other user-specific useful things for the UI in the profile.

### DIFF
--- a/db/migrations/20210308070729-add-captain-to-group.js
+++ b/db/migrations/20210308070729-add-captain-to-group.js
@@ -2,9 +2,6 @@
 
 module.exports = {
   up: async (queryInterface, Sequelize) => {
-    await queryInterface.bulkDelete('Shipments', {})
-    await queryInterface.bulkDelete('Groups', {})
-
     await queryInterface.addColumn('Groups', 'captainId', {
       allowNull: false,
       type: Sequelize.INTEGER,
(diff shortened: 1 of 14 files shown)


db/migrations/20210308070729-add-captain-to-group.js
```diff
@@ -0,0 +1,18 @@
+'use strict'
+
+module.exports = {
+  up: async (queryInterface, Sequelize) => {
+    await queryInterface.bulkDelete('Shipments', {})
+    await queryInterface.bulkDelete('Groups', {})
```
deammer: Question: why is this necessary?

src/apolloServer.ts
```diff
@@ -7,17 +7,25 @@ import depthLimit from 'graphql-depth-limit'
...
+} from './authenticateRequest'
 
 export type Context = {
   auth: Auth
 }
 
+export type AuthenticatedContext = {
+  auth: AuthenticatedAuth
+}
+
```
deammer, line 23: 💯

deammer (approved):
Looks good to me!

db/migrations/20210308070729-add-captain-to-group.js
bion: Eh, it probably isn't. Everyone is going to need to reset their DBs anyway. I'll remove it.

PR #78, opened 2021-03-12
Fix auth for local development

bion:
Without the proxy configured in the react dev server, the dev server would serve the react app for GET requests to `/profile`. This breaks auth for local development when running the react dev server because fetching `/profile` creates a user account for the authenticated user if they don't have one yet.

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -51,5 +51,6 @@
       "last 1 firefox version",
       "last 1 safari version"
     ]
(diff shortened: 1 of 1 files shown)


deammer (approved):

deammer:
That's totally my bad, I removed it in an ealier PR because I'd overlooked that endpoint. Thank you for fixing it!

Issue #69, opened 2021-03-09
Extract graphql queries, mutations, and fragments into their own files


deammer:
Suggestion (non-blocking): it'd be nice to pull these into their own files e.g. https://github.com/distributeaid/toolbox/blob/saga/react_ui/src/pages/createChapter.graphql

_Originally posted by @bion in https://github.com/distributeaid/shipment-tracker/pull/65#discussion_r589978463_

PR #81, opened 2021-03-14
Added a page to edit existing shipments

### LINKED ISSUE #10: Admin UI for group creation and editing
Background: the legacy react front-end (see background in #5) has a form for creating groups, but the shape of the data is different from what is needed for this project. I suggest taking a look at that code and seeing if it's worth migrating it over or writing this feature from scratch.

Done when: admins can create and edit groups using the UI

deammer:
This is half of the work required in #10. Creating shipments in the UI will come later.

### Changes

- added a new page to edit existing shipments
- adjusted the `TextInput` and `SelectInput` components so that they could output numbers (GraphQL is very strict about validations)
- added a shipment name formatter

### Screenshots

Rows in this table:

Screen Shot 2021-03-14 at 10 30 20 AM

...link to this page:

Screen Shot 2021-03-13 at 6 20 46 PM

### DIFF
--- a/frontend/src/pages/shipments/ShipmentEditPage.tsx
+++ b/frontend/src/pages/shipments/ShipmentEditPage.tsx
@@ -1,0 +1,86 @@
+import { gql, useMutation, useQuery } from '@apollo/client'
+import { FunctionComponent } from 'react'
+import { useParams } from 'react-router-dom'
+import LayoutWithNav from '../../layouts/LayoutWithNav'
+import { Shipment, ShipmentUpdateInput } from '../../types/api-types'
+import { formatShipmentName } from '../../utils/format'
+import ShipmentForm from './ShipmentForm'
+
+const ALL_SHIPMENT_FIELDS = gql`
+  fragment AllShipmentFields on Shipment {
+    id
+    shippingRoute
+    labelYear
+    labelMonth
+    offerSubmissionDeadline
+    status
+  }
+`
+
+const GET_SHIPMENT = gql`
+  ${ALL_SHIPMENT_FIELDS}
+  query Shipment($id: Int!) {
+    shipment(id: $id) {
+      ...AllShipmentFields
+    }
+  }
+`
+
+const UPDATE_SHIPMENT_MUTATION = gql`
+  ${ALL_SHIPMENT_FIELDS}
+  mutation Shipment($id: Int!, $input: ShipmentUpdateInput!) {
+    updateShipment(id: $id, input: $input) {
+      ...AllShipmentFields
+    }
+  }
+`
+
+const ShipmentEditPage: FunctionComponent = () => {
+  // Extract the shipment's ID from the URL
+  const { shipmentId } = useParams<{ shipmentId: string }>()
+
+  // Load the shipment's information
+  const { data: originalShipmentData, loading: queryIsLoading } = useQuery<{
+    shipment: Shipment
+  }>(GET_SHIPMENT, {
+    variables: { id: parseInt(shipmentId, 10) },
+  })
+
+  // Set up the mutation to update the shipment
+  const [updateShipment, { loading: mutationIsLoading }] = useMutation(
+    UPDATE_SHIPMENT_MUTATION,
+  )
+
+  const onSubmit = (input: ShipmentUpdateInput) => {
+    updateShipment({
+      variables: { id: parseInt(shipmentId, 10), input },
+    }).catch((error) => {
+      console.log(error)
+    })
+  }
+
+  return (
+    <LayoutWithNav>
+      <div className="max-w-5xl mx-auto border-l border-r border-gray-200 min-h-content">
+        <header className="p-4 md:p-6 border-b border-gray-200">
+          <h1 className="text-navy-800 text-3xl mb-2">
+            {originalShipmentData
+              ? formatShipmentName(originalShipmentData.shipment)
+              : 'Shipment'}
+          </h1>
+          <p className="text-gray-700">TODO describe what a shipment is</p>
+        </header>
+        <main className="p-4 md:p-6 max-w-lg pb-20">
+          <ShipmentForm
+            isLoading={queryIsLoading || mutationIsLoading}
+            submitButtonLabel="Save changes"
+            onSubmit={onSubmit}
+            defaultValues={originalShipmentData?.shipment}
+          />
+        </main>
+      </div>
+    </LayoutWithNav>
+  )
+}
+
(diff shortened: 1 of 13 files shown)


frontend/src/pages/shipments/ShipmentEditPage.tsx
```diff
@@ -0,0 +1,86 @@
...
+`
+
+const UPDATE_SHIPMENT_MUTATION = gql`
+  ${ALL_SHIPMENT_FIELDS}
+  mutation Shipment($id: Int!, $input: ShipmentUpdateInput!) {
+    updateShipment(id: $id, input: $input) {
+      ...AllShipmentFields
+    }
+  }
+`
```
deammer, line 36: I'll move these out to their own files in a future PR (#69). It'll require some changes to our codegen stuff and I'd like to keep that isolated.

frontend/src/pages/shipments/ShipmentForm.tsx
```diff
@@ -0,0 +1,169 @@
...
+  { label: 'Complete', value: ShipmentStatus.Complete },
+  { label: 'In progress', value: ShipmentStatus.InProgress },
+  { label: 'Open', value: ShipmentStatus.Open },
+  { label: 'In staging', value: ShipmentStatus.Staging },
+]
+
+const SHIPPING_ROUTE_OPTIONS = Object.keys(ShippingRoute).map((routeKey) => ({
+  label: (ShippingRoute as any)[routeKey],
+  value: (ShippingRoute as any)[routeKey],
+}))
```
deammer: I couldn't come up with a more concise way to transform an enum into an object. This works I guess 🙈

frontend/src/pages/shipments/ShipmentForm.tsx
```diff
@@ -0,0 +1,169 @@
...
+
+const GET_ALL_GROUPS = gql`
+  query Groups {
+    listGroups {
+      id
+      name
+      groupType
+    }
```
deammer, line 58: Finally starting to see the upside of GraphQL: only fetching the fields we need, and nothing else.

frontend/src/pages/shipments/ShipmentList.tsx
```diff
@@ -33,27 +35,26 @@ const SHIPMENTS_QUERY = gql`
   }
 `
 
-const COLUMNS = [
+const COLUMNS: Column<Shipment>[] = [
   {
-    Header: 'ID',
-    accessor: 'id',
+    Header: 'Name',
+    accessor: (row) => formatShipmentName(row),
```
deammer, line 42: This is something like `UK-2021-01` instead of the shipment's ID

frontend/src/pages/shipments/ShipmentForm.tsx
bion: Yeah I'm still waiting to see the benefits, it's still pretty unfamiliar. I guess the main positive thing seems to be standardized type definitions for everything which I've never tried to do with a REST API. I'm too used to protocol buffers at this point.

frontend/src/pages/shipments/ShipmentForm.tsx
deammer: I could simplify to:
```ts
const SHIPPING_ROUTE_OPTIONS = Object.keys(ShippingRoute).map((routeKey) => ({
  label: routeKey,
  value: routeKey,
}))
```

andyjac: Not sure if this is the proper way to do this, but we could create a helper method for getting enum keys (full context in [this blog post](https://www.petermorlion.com/iterating-a-typescript-enum/)):

```typescript
function enumKeys<O extends object, K extends keyof O = keyof O>(obj: O): K[] {
  return Object.keys(obj) as K[]
}

const SHIPPING_ROUTE_OPTIONS = enumKeys(ShippingRoute).map((routeKey) => ({
  label: ShippingRoute[routeKey],
  value: ShippingRoute[routeKey],
}))
```

This works surprisingly well and fully understands the typing of `routeKey` in context. For example, if I use an invalid key I get a proper error:

Screen Shot 2021-03-14 at 11 13 29 AM

bion: Suggestion (blocking): this seems like a great thing to add!

frontend/src/pages/shipments/ShipmentList.tsx
```diff
@@ -115,10 +116,18 @@ const ShipmentList: FunctionComponent = () => {
...
                         {...cell.getCellProps()}
                         className={cx('p-2 first:pl-6 last:pr-6', {
+                          'font-semibold text-navy-700 hover:underline':
+                            cell.column.Header === 'Name',
                           'bg-gray-50': cell.column.isSorted,
                         })}
                       >
-                        {cell.render('Cell')}
+                        {cell.column.Header === 'Name' ? (
+                          <Link to={`/shipment/${row.original.id}`}>
```
bion: Suggestion (blocking): recommend adding a function to routes.ts `shipmentShow(id: number)` that generates this string

frontend/src/utils/routes.ts
```diff
@@ -6,6 +6,7 @@ const ROUTES = {
   GROUP_CREATE: '/group/new',
   GROUP_EDIT: '/group/:groupId',
   SHIPMENT_LIST: '/shipments',
+  SHIPMENT_EDIT: '/shipment/:shipmentId',
```
bion: Suggestion (blocking): this URL path looks like a path for viewing the details of a shipment rather than editing one. Recommend changing to `/shipment/:id/edit`

frontend/src/pages/shipments/ShipmentForm.tsx
```diff
@@ -0,0 +1,169 @@
...
+    },
+    [hubListIsLoading, hubs],
+  )
+
+  const { register, handleSubmit, reset } = useForm()
+
+  useEffect(
+    function resetFormValues() {
+      if (props.defaultValues) {
+        console.log(props.defaultValues)
```
bion: Question (blocking): log statement left in intentionally?

frontend/src/pages/shipments/ShipmentForm.tsx
deammer: Thanks Andrew!

frontend/src/pages/shipments/ShipmentList.tsx
deammer: I like it, done

bion (approved):

PR #84, opened 2021-03-14
Implement offer queries

### LINKED ISSUE #22: Offer query API
Background: see `Offer` type in proposed [GraphQL schema here](https://www.notion.so/distributeaid/Technical-requirements-c2fd190e0a8d4f708119c6944fa654dd#ee90467d687d4d158a5a97d131c7a862).

This **does not include** the API work for offer sub-components `Pallet` and `LineItem`.

Done when graphql API supports listing all offers for a specific shipment and querying a specific offer by ID.

Depends on #21

bion:
Closes #22

### DIFF
--- a/src/tests/offers_api.test.ts
+++ b/src/tests/offers_api.test.ts
@@ -12,7 +12,7 @@ import {
   ShipmentStatus,
   ShippingRoute,
 } from '../server-internal-types'
-import { makeTestServer } from '../testServer'
+import { makeAdminTestServer, makeTestServer } from '../testServer'
 import { createGroup, createShipment } from './helpers'
 import UserAccount from '../models/user_account'
 import { fakeUserAuth } from '../authenticateRequest'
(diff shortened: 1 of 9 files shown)


src/tests/offers_api.test.ts
```diff
@@ -319,4 +320,165 @@ describe('Offers API', () => {
...
+      }
+    `
+
+    beforeEach(async () => {
+      otherShipment = await createShipment({
+        labelYear: 2021,
+        labelMonth: 1,
+        shippingRoute: ShippingRoute.Uk,
+        sendingHubId: captainsGroup.id,
+        receivingHubId: captainsGroup.id,
```
deammer, line 352: Question: would this be possible in the "real world"? Having the same sending and receiving hub? I wonder if we should validate that on the backend. On the frontend, we split shipping groups by `groupType` so that this can never happen.

deammer (approved):
🚢

src/tests/offers_api.test.ts
bion: There's nothing preventing it from happening in the real world if an admin decides to do it. Since only admins can mutate shipments though I'm less worried about locking this down with rigorous consistency checks.

PR #91, opened 2021-03-15
Factor out common offer and pallet authorization logic

### LINKED ISSUE #24: Pallet query API
Background: see `Pallet` type in proposed [GraphQL schema here](https://www.notion.so/distributeaid/Technical-requirements-c2fd190e0a8d4f708119c6944fa654dd#ee90467d687d4d158a5a97d131c7a862).

This **does not include** the API work for the sub-component `LineItem`.

Done when graphql API supports listing all pallets for a given offer as well as all pallets for a given shipment, as well as querying for a specific pallet by ID.

Depends on #23

bion:
Introduces two new functions `authorizeOfferMutation(offer: Offer, context: AuthenticatedContext): void` and `authorizeOfferQuery(offer: Offer, context: AuthenticatedContext): void`

### DIFF
--- a/src/resolvers/offer/offer_authorization.ts
+++ b/src/resolvers/offer/offer_authorization.ts
@@ -3,6 +3,12 @@ import { AuthenticatedContext } from '../../apolloServer'
 import Offer from '../../models/offer'
 import { OfferStatus, ShipmentStatus } from '../../server-internal-types'
 
+/**
+ * Asserts that an offer has a shipment and sending group.
+ * Throws {@link ApolloError} if missing either
+ *
+ * @param {Offer} offer - the offer check consistency of
+ * */
 const assertAssociations = (offer: Offer): void => {
   if (!offer.sendingGroup) {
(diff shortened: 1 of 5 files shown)


src/resolvers/offer/offer_authorization.ts
```diff
@@ -0,0 +1,56 @@
+import { ApolloError, ForbiddenError } from 'apollo-server-express'
+import { AuthenticatedContext } from '../../apolloServer'
+import Offer from '../../models/offer'
+import { OfferStatus, ShipmentStatus } from '../../server-internal-types'
+
+const assertAssociations = (offer: Offer): void => {
```
deammer, line 12: Suggestion (non-blocking): some JSDoc-style comments for these methods would be very helpful!

deammer (approved):
This is a super clean approach, I like it 👍

src/resolvers/offer/offer_authorization.ts
bion: done!

PR #96, opened 2021-03-16
Line item mutation resolvers

### LINKED ISSUE #25: LineItem mutate API
Background: see `LineItem` type in proposed [GraphQL schema here](https://www.notion.so/distributeaid/Technical-requirements-c2fd190e0a8d4f708119c6944fa654dd#ee90467d687d4d158a5a97d131c7a862).

Done when graphql API supports creating and updating `LineItems` in specific `Pallets`.

Depends on #24

bion:
Implements mutation APIs for:

* creating line items
* updating line items
* destroying line items
* moving line items from one pallet to another in the same offer

Also adds some additional validation logic for setting enum values in all resolvers with enums.

This is fairly light on testing relative to the other resources we've implemented so far, but it has a test for the success-path for each mutation. Validation and authorization logic isn't tested for the most part, though the latter is mostly reusing code that is tested heavily in other resolves.

closes #25


schema.graphql
```diff
@@ -91,6 +91,10 @@ type Mutation {
   addPallet(input: PalletCreateInput!): Pallet!
   updatePallet(id: Int!, input: PalletUpdateInput!): Pallet!
   destroyPallet(id: Int!): Offer! # return the destroyed pallet's offer
+  addLineItem(palletId: Int!): Pallet! # return the pallet the line item was added to
```
deammer, line 94: Thought: returning a `Pallet` feels weird because I would normally expect a `LineItem`. What's a use case for this?

bion: The line item is displayed in the context of the pallet as an entry in its list of line items. I figured returning the pallet would allow the mutation caller to specify a query that includes the list of line items for that pallet so upon success can update the client-side pallet's line item list in entirety rather than attempting to book-keep adding the new line item. My guess is that would make things simpler on the client. Another plus would be if two people were editing the same offer at the same time their clients would be less likely to get out of sync.

src/resolvers/validateEnumMembership.ts
```diff
@@ -0,0 +1,17 @@
+import { UserInputError } from 'apollo-server'
+import { difference } from 'lodash'
```
deammer, line 2: Nit (non-blocking): this is a tiny optimization but this import statement will add all of lodash's modules to the bundle. You could do this instead:

```ts
import difference from 'lodash/difference'
```

https://selleo.com/til/posts/0ac258eyv8-decrease-bundle-size-by-importing-lodash-correctly

schema.graphql
bion: Based on what I'm reading in the [blog post you linked](https://www.apollographql.com/blog/when-to-use-refetch-queries-in-apollo-client/), it sounds like this behavior could make things easier with managing the cache:
> One way we could omit needing to write this update function is if the mutation response returned the entire list of new todos in the mutation response, but that’s not always possible.

deammer (approved):
Looking good

src/resolvers/validateEnumMembership.ts
bion: We're not building a bundle for server-side code so wouldn't this have no effect?

schema.graphql
deammer: Those are very good reasons 👍

src/resolvers/validateEnumMembership.ts
deammer: Okay yeah I need some sleep 🤦

bion: 😄 well-earned, thanks for the review!

PR #97, opened 2021-03-18
Update the list of entities after creating new ones

### LINKED ISSUE #95: The client-side cache doesn't refresh when creating entities
In #94, we discovered that when a new shipment is created, the list of shipments is not refreshed. We need to figure out how to fix it and document the best practices!

### Repro steps

1. View the list of shipments
2. Create a new shipment
3. Go back to the list of shipments

**Expected result:** the new shipment is displayed in the list

**Actual result:** the new shipment is absent from the list

deammer:
This fixes #95, where the lists of groups and shipments don't get updated after creating new groups or shipments.

To my understanding, there are 2 ways to refresh the Apollo cache:
- call the "list all things" query again (downside: 2 queries instead of 1 when creating a new entity), or
- update the cache manually (downside: way more complicated than it should be)

I decided to go with the first method because it's easy to understand. As we become more familiar with Apollo, we can revisit this if necessary. [This was a helpful resource](https://www.apollographql.com/blog/when-to-use-refetch-queries-in-apollo-client/).

### DIFF
--- a/frontend/src/pages/shipments/ShipmentCreatePage.tsx
+++ b/frontend/src/pages/shipments/ShipmentCreatePage.tsx
@@ -2,6 +2,7 @@ import { FunctionComponent } from 'react'
 import { useHistory } from 'react-router-dom'
 import LayoutWithNav from '../../layouts/LayoutWithNav'
 import {
+  AllShipmentsDocument,
   ShipmentCreateInput,
   useCreateShipmentMutation,
 } from '../../types/api-types'
@@ -18,7 +19,11 @@ const ShipmentCreatePage: FunctionComponent = () => {
 
   const onSubmit = (input: ShipmentCreateInput) => {
     // Create the shipment and then redirect to its "view" page
-    addShipment({ variables: { input } })
+    addShipment({
+      variables: { input },
+      // Fetch the updated list of shipments
+      refetchQueries: [{ query: AllShipmentsDocument }],
+    })
       .then(({ data }) => {
         if (data) {
           const newShipmentId = data.addShipment.id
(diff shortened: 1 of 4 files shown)


frontend/src/pages/shipments/ShipmentCreatePage.tsx
```diff
@@ -18,7 +19,11 @@ const ShipmentCreatePage: FunctionComponent = () => {
 
   const onSubmit = (input: ShipmentCreateInput) => {
     // Create the shipment and then redirect to its "view" page
-    addShipment({ variables: { input } })
+    addShipment({
+      variables: { input },
+      // Fetch the updated list of shipments
+      refetchQueries: [{ query: AllShipmentsDocument }],
+    })
```
bion, line 26: Question (non-blocking): am I reading the docs right that if the `addShipment` mutation returned a list of all shipments with their IDs we wouldn't need a refetch query *or* update function?

bion (approved):
Thanks for the blog post link. Yeah this seems like a quick way to keep moving for now. Getting our heads around cache control and query refetching sooner than later seems like it'll be good though. Sounds like knowing about these things will be super handy when we start to think about an offline mode.

frontend/src/pages/shipments/ShipmentCreatePage.tsx
deammer: Yeah that seems to be the case but I feel a little uncomfortable with modifying the API in order to address a client-side implementation detail. I think I'll revisit this when I have a little more brain power.

bion: Glad to be learning about this stuff! I'm happy to adjust the API as we learn what will be most convenient for the UI.

PR #98, opened 2021-03-18
Remove nested directories in src/resolvers

bion:
The resolvers for each resource were in their own subdir of `src/resolvers`. I've found this to be slightly annoying with fuzzy finding and file management since there end up being a bunch of open files with the same basename of `index.ts`. This PR removes the subdirs by making the following moves:

```
src/resolvers/{group/index.ts => group.ts}
src/resolvers/{line_items/index.ts => line_items.ts}
src/resolvers/{offer/index.ts => offer.ts}
src/resolvers/{offer => }/offer_authorization.ts
src/resolvers/{pallet/index.ts => pallet.ts}
src/resolvers/{shipment/index.ts => shipment.ts}
```

### DIFF
--- a/src/resolvers/group.ts
+++ b/src/resolvers/group.ts
@@ -1,13 +1,13 @@
 import { ApolloError, ForbiddenError, UserInputError } from 'apollo-server'
 import { has } from 'lodash'
-import Group, { GroupAttributes } from '../../models/group'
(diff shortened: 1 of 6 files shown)


andyjac (approved):
Much better! Thank you 👍

Issue #34, opened 2021-02-23
Sending group leaders can create an offer for an active shipment


bion:
depends on #33

deammer:
Implemented in #101

PR #101, opened 2021-03-19
Added a page to create offers

deammer:
Related to #34 

### Changes

- added a page to create offers
- added a page to view a single offer (it's very very basic)
- wrapped the `/profile` call into its own component. The profile data is accessed using React's context API
- added a list of offers to the shipment page

### Screen recording

https://user-images.githubusercontent.com/3411183/111850653-1d3b5400-88ce-11eb-851d-8e7919c95ceb.mov

### DIFF
--- a/frontend/src/AppRoot.tsx
+++ b/frontend/src/AppRoot.tsx
@@ -1,8 +1,9 @@
 import { useAuth0 } from '@auth0/auth0-react'
-import React, { useEffect, useState } from 'react'
+import React from 'react'
 import { BrowserRouter as Router, Route, Switch } from 'react-router-dom'
 import ApolloAuthProvider from './components/ApolloAuthProvider'
 import PrivateRoute from './components/PrivateRoute'
+import { UserProfileProvider } from './components/UserProfileContext'
 import AdminPage from './pages/AdminPage'
 import ApolloDemoPage from './pages/ApolloDemo'
 import GroupCreatePage from './pages/groups/GroupCreatePage'
@@ -13,6 +14,8 @@ import HomePage from './pages/Home'
 import KitchenSink from './pages/KitchenSink'
 import LoadingPage from './pages/LoadingPage'
 import NotFoundPage from './pages/NotFoundPage'
+import CreateOfferPage from './pages/offers/CreateOfferPage'
+import ViewOfferPage from './pages/offers/ViewOfferPage'
 import PublicHomePage from './pages/PublicHome'
 import ShipmentCreatePage from './pages/shipments/ShipmentCreatePage'
 import ShipmentEditPage from './pages/shipments/ShipmentEditPage'
@@ -22,82 +25,69 @@ import ROUTES from './utils/routes'
 
 const isDev = process.env.NODE_ENV === 'development'
 
-const fetchProfile = (token: string) => {
-  return fetch('/profile', {
-    headers: { Authorization: `Bearer ${token}` },
-  })
-}
-
 const AppRoot = () => {
-  const { isAuthenticated, isLoading, getAccessTokenSilently } = useAuth0()
-  const [profileIsLoaded, setProfileIsLoaded] = useState(false)
-
-  useEffect(() => {
-    if (isAuthenticated && !profileIsLoaded) {
-      getAccessTokenSilently()
-        .then(fetchProfile)
-        .then((response) => {
-          if (response.ok) {
-            setProfileIsLoaded(true)
-          } else {
-            console.error('Non-OK server response retrieving profile')
-          }
-        })
-    }
-  }, [isAuthenticated, getAccessTokenSilently, profileIsLoaded])
+  const { isLoading, isAuthenticated } = useAuth0()
 
   return (
     <ApolloAuthProvider>
-      <Router>
-        <Switch>
-          {isDev && (
-            <Route path={ROUTES.KITCHEN_SINK}>
-              <KitchenSink />
+      <UserProfileProvider>
+        <Router>
+          <Switch>
+            {isDev && (
+              <Route path={ROUTES.KITCHEN_SINK}>
+                <KitchenSink />
+              </Route>
+            )}
+            {isLoading && (
+              <Route>
+                <LoadingPage />
+              </Route>
+            )}
+            <Route path={ROUTES.HOME} exact>
+              {isAuthenticated ? <HomePage /> : <PublicHomePage />}
             </Route>
-          )}
-          {isLoading && (
-            <Route>
-              <LoadingPage />
+            <PrivateRoute path={ROUTES.ADMIN_ROOT} exact>
+              <AdminPage />
+            </PrivateRoute>
+            <Route path={ROUTES.APOLLO_DEMO}>
+              <ApolloDemoPage />
             </Route>
-          )}
-          <Route path={ROUTES.HOME} exact>
-            {isAuthenticated ? <HomePage /> : <PublicHomePage />}
-          </Route>
-          <PrivateRoute path={ROUTES.ADMIN_ROOT} exact>
-            <AdminPage />
-          </PrivateRoute>
-          <Route path={ROUTES.APOLLO_DEMO}>
-            <ApolloDemoPage />
-          </Route>
-          <PrivateRoute path={ROUTES.GROUP_LIST}>
-            <GroupList />
-          </PrivateRoute>
-          <PrivateRoute path={ROUTES.GROUP_CREATE}>
-            <GroupCreatePage />
-          </PrivateRoute>
-          <PrivateRoute path={ROUTES.GROUP_EDIT}>
-            <GroupEditPage />
-          </PrivateRoute>
-          <PrivateRoute path={ROUTES.GROUP_VIEW}>
-            <GroupViewPage />
-          </PrivateRoute>
-          <PrivateRoute path={ROUTES.SHIPMENT_LIST}>
-            <ShipmentList />
-          </PrivateRoute>
-          <PrivateRoute path={ROUTES.SHIPMENT_CREATE}>
-            <ShipmentCreatePage />
-          </PrivateRoute>
-          <PrivateRoute path={ROUTES.SHIPMENT_EDIT}>
-            <ShipmentEditPage />
-          </PrivateRoute>
-          <PrivateRoute path={ROUTES.SHIPMENT_VIEW}>
-            <ShipmentViewPage />
-          </PrivateRoute>
-          <PrivateRoute path="*">
-            <NotFoundPage />
-          </PrivateRoute>
-        </Switch>
-      </Router>
+            <PrivateRoute path={ROUTES.GROUP_LIST}>
+              <GroupList />
+            </PrivateRoute>
+            <PrivateRoute path={ROUTES.GROUP_CREATE}>
+              <GroupCreatePage />
+            </PrivateRoute>
+            <PrivateRoute path={ROUTES.GROUP_EDIT}>
+              <GroupEditPage />
+            </PrivateRoute>
+            <PrivateRoute path={ROUTES.GROUP_VIEW}>
+              <GroupViewPage />
+            </PrivateRoute>
+            <PrivateRoute path={ROUTES.SHIPMENT_OFFER_CREATE}>
+              <CreateOfferPage />
+            </PrivateRoute>
+            <PrivateRoute path={ROUTES.SHIPMENT_OFFER_VIEW}>
+              <ViewOfferPage />
+            </PrivateRoute>
+            <PrivateRoute path={ROUTES.SHIPMENT_LIST}>
+              <ShipmentList />
+            </PrivateRoute>
+            <PrivateRoute path={ROUTES.SHIPMENT_CREATE}>
+              <ShipmentCreatePage />
+            </PrivateRoute>
+            <PrivateRoute path={ROUTES.SHIPMENT_EDIT}>
+              <ShipmentEditPage />
+            </PrivateRoute>
+            <PrivateRoute path={ROUTES.SHIPMENT_VIEW}>
+              <ShipmentViewPage />
+            </PrivateRoute>
+            <PrivateRoute path="*">
+              <NotFoundPage />
+            </PrivateRoute>
+          </Switch>
+        </Router>
+      </UserProfileProvider>
     </ApolloAuthProvider>
   )
 }
--- a/frontend/src/AppRoot.tsx
+++ b/frontend/src/AppRoot.tsx
@@ -1,5 +1,4 @@
 import { useAuth0 } from '@auth0/auth0-react'
-import React from 'react'
 import { BrowserRouter as Router, Route, Switch } from 'react-router-dom'
 import ApolloAuthProvider from './components/ApolloAuthProvider'
 import PrivateRoute from './components/PrivateRoute'
--- a/frontend/src/components/UserProfileContext.tsx
+++ b/frontend/src/components/UserProfileContext.tsx
@@ -7,8 +7,6 @@ export interface UserProfile {
 }
 
 const fetchProfile = (token: string) => {
-  console.log("fetching the user's profile")
-
   return fetch('/profile', {
     headers: { Authorization: `Bearer ${token}` },
   })
--- a/frontend/src/components/UserProfileContext.tsx
+++ b/frontend/src/components/UserProfileContext.tsx
@@ -1,0 +1,50 @@
+import { useAuth0 } from '@auth0/auth0-react'
+import { createContext, FunctionComponent, useEffect, useState } from 'react'
+
+export interface UserProfile {
+  id: number
+  isAdmin: boolean
+}
+
+const fetchProfile = (token: string) => {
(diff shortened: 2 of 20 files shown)


frontend/src/AppRoot.tsx
```diff
@@ -22,82 +25,69 @@ import ROUTES from './utils/routes'
...
+  const { isLoading, isAuthenticated } = useAuth0()
 
   return (
     <ApolloAuthProvider>
-      <Router>
-        <Switch>
-          {isDev && (
-            <Route path={ROUTES.KITCHEN_SINK}>
-              <KitchenSink />
+      <UserProfileProvider>
```
deammer, line 32: I wrapped the `/profile` call into its own Context component. This way, it's isolated into its own file, and we can access the user's profile from anywhere in the app.

frontend/src/components/UserProfileContext.tsx
```diff
@@ -0,0 +1,50 @@
...
+            console.error('Non-OK server response retrieving profile')
+          }
+
+          return response.json()
+        })
+        .then((data) => {
+          setProfile(data)
+        })
+    }
+  }, [isAuthenticated, getAccessTokenSilently, profileIsLoaded])
```
deammer, line 39: This code was copied over from AppRoot, except that we're not storing the user's profile. This way, any component nested within the provider can access that information without an extra call to the backend.

frontend/src/components/tabs/TabLink.tsx
```diff
@@ -0,0 +1,24 @@
+import cx from 'classnames'
+import { FunctionComponent } from 'react'
+import { Link, LinkProps, useLocation } from 'react-router-dom'
+
+const TabLink: FunctionComponent<LinkProps> = ({ children, ...otherProps }) => {
```
deammer, line 5: Added the `TabLink` and `TabList` components. They are integrated with React Router, which means they're already in sync with the URL.

Screen Shot 2021-03-19 at 3 36 48 PM

frontend/src/pages/offers/CreateOfferForm.tsx
```diff
@@ -0,0 +1,124 @@
...
+
+  const groupForUser = useMemo(() => {
+    if (groups?.listGroups && profile) {
+      return groups.listGroups.find(
+        (group) =>
+          group.captainId === profile.id &&
+          group.groupType === GroupType.SendingGroup,
+      )
+    }
+  }, [groups, profile])
```
deammer, line 49: Here we start to get into the weeds because we need data from multiple endpoints.

We want to display some information about the group the user is a captain of. To do that, we fetch the list of all the groups and figure out which one has a matching `captainId`. It would be better if we could query a group by captain ID in the future.

frontend/src/pages/offers/CreateOfferForm.tsx
```diff
@@ -0,0 +1,124 @@
...
+
+/**
+ * This form allows users to create a new offer. It is very specific to offer
+ * creation, and therefore difficult to reuse for editing offers.
+ */
+const OfferForm: FunctionComponent<Props> = (props) => {
+  // Each group leader is assigned to a single group
+  // We need to figure out which group that is
+  // TODO: handle the fact that admins can be assigned to any number of groups
+  const profile = useContext(UserProfileContext)
```
deammer, line 38: Here's where we consume the user's profile.

frontend/src/pages/shipments/ShipmentDetails.tsx
```diff
@@ -0,0 +1,86 @@
...
+  formatCountryCodeToName,
+  formatLabelMonth,
+  getShipmentStatusBadgeColor,
+} from '../../utils/format'
+
+interface Props {
+  shipmentId: number
+}
+
+const ShipmentDetails: FunctionComponent<Props> = ({ shipmentId }) => {
```
deammer, line 16: This area already existing but it was moved into a tab. Here's what it looks like:

Screen Shot 2021-03-19 at 3 45 46 PM

frontend/src/pages/shipments/ShipmentOffers.tsx
```diff
@@ -0,0 +1,121 @@
...
+  useAllGroupsMinimalQuery,
+  useOffersForShipmentQuery,
+} from '../../types/api-types'
+import { offerCreateRoute } from '../../utils/routes'
+
+interface Props {
+  shipmentId: number
+}
+
+const ShipmentOffers: FunctionComponent<Props> = ({ shipmentId }) => {
```
deammer, line 18: This is the Offers tabs:

Screen Shot 2021-03-19 at 3 46 43 PM

frontend/src/utils/routes.ts
```diff
@@ -9,24 +11,42 @@ const ROUTES = {
...
+  SHIPMENT_OFFER_LIST: '/shipment/:shipmentId/offers',
+  SHIPMENT_OFFER_CREATE: '/shipment/:shipmentId/offer/new',
+  SHIPMENT_OFFER_VIEW: '/shipment/:shipmentId/offer/:offerId',
   SHIPMENT_EDIT: '/shipment/:shipmentId/edit',
   KITCHEN_SINK: '/kitchen-sink',
 }
 
 export function groupViewRoute(groupId: number | string) {
-  return ROUTES.GROUP_VIEW.replace(':groupId', groupId.toString())
+  return generatePath(ROUTES.GROUP_VIEW, { groupId })
```
deammer, line 22: I realized that `react-router` has a neat method to generate a path from a route.

```ts
generatePath("/shipment/:shipmentId", { shipmentId: "1" })
// -> "/shipment/1"
```

frontend/src/components/UserProfileContext.tsx
bion: Nice, yet another react feature for me to learn 😄

frontend/src/components/tabs/TabLink.tsx
bion: ah nice!

frontend/src/data/constants.ts
```diff
@@ -1,4 +1,5 @@
-import { GroupType } from '../types/api-types'
+import { GroupType, OfferStatus } from '../types/api-types'
+import { enumValues } from '../utils/types'
```
bion, line 2: thought (non-blocking): out of scope for this PR but i'd like to pull the frontend utils folder into the root of the repo for sharing with the backend.

frontend/src/pages/offers/CreateOfferForm.tsx
```diff
@@ -0,0 +1,118 @@
...
+}
+
+/**
+ * This form allows users to create a new offer. It is very specific to offer
+ * creation, and therefore difficult to reuse for editing offers.
+ */
+const OfferForm: FunctionComponent<Props> = (props) => {
+  // Each group leader is assigned to a single group
+  // We need to figure out which group that is
+  // TODO: handle the fact that admins can be assigned to any number of groups
```
bion, line 37: many user accounts per group relationships is something to chat about

frontend/src/pages/offers/CreateOfferForm.tsx
bion: How about an endpoint that's like `listEditableGroups()` that returns all the groups the user has edit access to? This would handle the admin and non-admin cases with a single endpoint.

frontend/src/pages/offers/CreateOfferForm.tsx
```diff
@@ -0,0 +1,118 @@
...
+          group.groupType === GroupType.SendingGroup,
+      )
+    }
+  }, [groups, profile])
+
+  const { data: targetShipment } = useShipmentQuery({
+    variables: { id: props.shipmentId },
+  })
+
+  // Set up the form
```
bion: nit (non-blocking): my very slight preference would be no comment here

frontend/src/pages/offers/CreateOfferForm.tsx
bion: In that case I guess we'd need to give the option of selecting the sending group for users with multiple 🤔

frontend/src/pages/offers/CreateOfferForm.tsx
```diff
@@ -0,0 +1,118 @@
...
+      <fieldset>
+        <legend>Photos</legend>
+      </fieldset>
+      <Button
+        variant="primary"
+        type="submit"
+        className="mt-6"
+        disabled={props.isLoading}
+      >
+        {props.submitButtonLabel}
```
bion: suggestion (non-blocking): since this form is specific to create, maybe it's ok to hard-code this?

frontend/src/pages/offers/ViewOfferPage.tsx
```diff
@@ -0,0 +1,61 @@
...
+  const params = useParams<{ shipmentId: string; offerId: string }>()
+  const offerId = parseInt(params.offerId, 10)
+  const shipmentId = parseInt(params.shipmentId, 10)
+
+  // Figure out the shipment ID from the path
+  const { data: targetShipment } = useShipmentQuery({
+    variables: { id: shipmentId },
+  })
+
+  const { data } = useOfferQuery({ variables: { id: offerId } })
```
bion, line 20: question (non-blocking): how should we handle 404s?

frontend/src/pages/offers/ViewOfferPage.tsx
```diff
@@ -0,0 +1,61 @@
...
+import ReadOnlyField from '../../components/forms/ReadOnlyField'
+import InternalLink from '../../components/InternalLink'
+import LayoutWithNav from '../../layouts/LayoutWithNav'
+import { useOfferQuery, useShipmentQuery } from '../../types/api-types'
+import { formatShipmentName } from '../../utils/format'
+import { shipmentViewOffersRoute } from '../../utils/routes'
+
+const ViewOfferPage: FunctionComponent = () => {
+  const params = useParams<{ shipmentId: string; offerId: string }>()
+  const offerId = parseInt(params.offerId, 10)
```
bion, line 12: question (non-blocking): how should we handle parse errors if someone typos `/offers/12w`?

frontend/src/utils/routes.ts
bion: ah nice!

frontend/src/pages/shipments/ShipmentDetails.tsx
bion: thought (non-blocking): maybe what we want is this view for the admin, and for sending group captains a view without the offers tab and instead a button to create an offer or a summary of the offer they've created with links to edit.

frontend/src/pages/shipments/ShipmentDetails.tsx
```diff
@@ -0,0 +1,86 @@
...
+  // Load the shipment's information
```
bion: suggestion (non-blocking): imo the comment doesn't add much

frontend/src/pages/shipments/ShipmentDetails.tsx
```diff
@@ -0,0 +1,86 @@
...
+            {shipmentData.sendingHub.name}
+          </div>
+          <div className="text-gray-600">
+            {shipmentData.sendingHub.primaryLocation.townCity},{' '}
+            {formatCountryCodeToName(
+              shipmentData.sendingHub.primaryLocation.countryCode,
+            )}
+          </div>
+        </div>
+        <div className="text-2xl text-gray-500 p-2 md:p-4 transform md:-rotate-90 text-center">
```
bion, line 65: `md:-rotate-90` slick

frontend/src/pages/shipments/ShipmentOffers.tsx
```diff
@@ -0,0 +1,131 @@
...
+  })
+
+  // Get the list of groups
+  const { data: groups } = useAllGroupsMinimalQuery()
+
+  const offers = useMemo(() => data?.listOffers || [], [data])
+
+  const COLUMNS = useMemo(() => {
+    if (groups?.listGroups && data?.listOffers) {
+      const columns: Column<OffersForShipmentQuery['listOffers'][0]>[] = [
```
bion: question (non-blocking): what's going on with the type here?

frontend/src/pages/shipments/ShipmentOffers.tsx
```diff
@@ -0,0 +1,131 @@
...
+        {
+          Header: 'Photos',
+          accessor: (row) => row.photoUris.length,
+        },
+      ]
+
+      return columns
+    }
+
+    return []
```
bion: nit (non-blocking): invert the boolean and early return instead of the large `if` body?

frontend/src/AppRoot.tsx
bion: ah that's a great improvement

bion (approved):

frontend/src/pages/offers/CreateOfferForm.tsx
deammer: And on top of that, if we move to multiple sending/receiving groups per shipment, we'll need to filter out the groups that aren't part of that shipment. So maybe something like:

```ts
const groupsFromShipment = shipment.sendingGroups.map(group => group.id)
const groupsToPickFrom = groups.listGroups.filter(
  group => 
    group.captainId === profile.id &&
    group.groupType = GroupType.SendingGroup &&
    groupsFromShipment.includes(group.id)
)
```

And then on the UI we can display a dropdown if there are 2 or more choices.

frontend/src/pages/shipments/ShipmentDetails.tsx
deammer: I didn't realize that group captains could only create a single offer per shipment. I totally agree with your thought then, let's simplify this for non-admins in a future PR.

frontend/src/data/constants.ts
deammer: I'd like that too, but it's not as straight-forward as it should be because `react-script` doesn't like importing code from outside the `/frontend` directory. I'm sure it's possible to hack around that though.

frontend/src/pages/offers/CreateOfferForm.tsx
deammer: Isn't it the opposite? Many groups per user account, right?

frontend/src/pages/offers/ViewOfferPage.tsx
deammer: Good point! #102

frontend/src/pages/offers/ViewOfferPage.tsx
deammer: I think that can be included in #102

frontend/src/pages/shipments/ShipmentOffers.tsx
deammer: Welcome to my hell! First, the point of the whole `Column<...>` type here is to get a nice API when defining table columns with `react-table`:
```ts
{
  Header: 'Pallets',
  // Here, TS knows that `row` is has a `pallets` property that's an array
  accessor: (row) => row.pallets.length,
}
```

So why does it look so gross? Because the codegen doesn't generate a _named type_ from the schema. To break it down:
```ts
Column<OffersForShipmentQuery['listOffers'][0]>
// ^            ^                  ^        ^
// 1            2                  3        4
```
1. `Column` comes from `react-table` and allows us to define type-safe... columns that are passed to the hook
2. `OfferForShipmentQuery` is the shape of the data that's returned from the `useOfferForShipmentQuery` hook:

```ts
type OffersForShipmentQuery = { __typename?: 'Query' } & {
  listOffers: Array<
    { __typename?: 'Offer' } & Pick<
      Offer,
      'id' | 'status' | 'sendingGroupId' | 'photoUris'
    > & { pallets: Array<{ __typename?: 'Pallet' } & Pick<Pallet, 'id'>> }
  >
}

// Extrapolated version:
type OffersForShipmentQuery = {
  __typename?: 'Query'
  listOffers: Array<{
    __typename?: 'Offer'
    id: number
    status: OfferStatus
    sendingGroupId: number
    photoUris: string[]
    pallets: Array<{
      __typename?: 'Pallet'
      id: number
    }>
  }
}
```

3. In the `OffersForShipmentQuery` type above, the `listOffers` property is what we care about: that's the array of offers we wanna display.
4. Finally, the `listOffers[0]` is what tells TS that we're looking for the shape of a single array element. It doesn't matter what the value of the index is.

**Can we do better?** Probably! Maybe there's a way to generate named types from `graphql-codegen`.

I've also tried relying on the `Offer` type but that's not compatible with the shape of the items in `OffersForShipmentQuery.listOffers`. We would have to pick the properties we care about and it would become even wordier.

frontend/src/pages/offers/CreateOfferForm.tsx
bion: many to many I'm afraid 😱

frontend/src/pages/shipments/ShipmentOffers.tsx
bion: Oh wow, thank you for the very clear explanation. Yeah it sounds like there's not a great option here. Nice work maintaining type safety and not just bypassing the checks!

PR #105, opened 2021-03-21
Built a modal system for the UI

deammer:
We're gonna need some modals so we can nest forms within forms within forms 😭 

### Changes

- added a modal system
- added the `body-scroll-lock` package which prevents users from scrolling the page when a modal is visible
- added the `react-transition-group` package for animations

https://user-images.githubusercontent.com/3411183/111924645-d337a700-8a62-11eb-965f-70038986e29c.mov


frontend/src/components/animation/TailwindTransition.tsx
```diff
@@ -0,0 +1,102 @@
+import { FunctionComponent } from 'react'
+import { CSSTransition as ReactCSSTransition } from 'react-transition-group'
```
bion, line 2: 🎶hello darkness my old friend🎶

frontend/src/components/animation/TailwindTransition.tsx
```diff
@@ -0,0 +1,102 @@
...
+  leaveFrom = '',
+  leaveTo = '',
+  ...otherProps
+}) => {
+  const enterClasses = enter.split(' ').filter((s) => s.length > 0)
+  const enterFromClasses = enterFrom.split(' ').filter((s) => s.length > 0)
+  const enterToClasses = enterTo.split(' ').filter((s) => s.length > 0)
+  const leaveClasses = leave.split(' ').filter((s) => s.length > 0)
+  const leaveFromClasses = leaveFrom.split(' ').filter((s) => s.length > 0)
+  const leaveToClasses = leaveTo.split(' ').filter((s) => s.length > 0)
```
bion: suggestion (blocking): please extract a method for `.split(' ').filter((s) => s.length > 0)` with a name that describes why it's necessary or a comment if clear name for that isn't easy

frontend/src/components/animation/TailwindTransition.tsx
```diff
@@ -0,0 +1,102 @@
...
+    if (classes.length > 0) {
+      node.classList.add(...classes)
+    }
+  }
+
+  const removeClasses = (node: HTMLElement, classes: string[]) => {
+    if (classes.length > 0) {
+      node.classList.remove(...classes)
+    }
+  }
```
bion: suggestion (blocking): `removeClasses` and `addClasses` don't access anything in the scope of the enclosing function so recommend declaring them outside of it.

frontend/src/components/model/BlankModal.tsx
```diff
@@ -0,0 +1,61 @@
...
+            <path
+              strokeLinecap="round"
+              strokeLinejoin="round"
+              strokeWidth={2}
+              d="M6 18L18 6M6 6l12 12"
+            />
+          </svg>
+        </button>
+      </div>
+      <div className="bg-white m-2 rounded-lg shadow-xl">{children}</div>
```
bion: suggestion (non-blocking): my read is that this component isn't so much blank as it is a wrapper that puts its children in a modal. consider renaming to something like `ModalWrapper` or even just `Modal`

bion: Ok woops just saw the next file

bion: What's this used for?

bion (approved):

frontend/src/components/model/BlankModal.tsx
deammer: Good point

deammer: Renamed `Modal` -> `ModalBase` and `BlankModal` -> `PlainModal`

PR #107, opened 2021-03-23
Add pricing to shipments

bion:
Also introduces a new type to the schema to represent quantities of money in any currency:
```graphql
type MoneyAmount {
  # Currency is a string holding an ISO-4217 currency code
  currency: Currency!
  quantityInMinorUnits: Int!
}
```

### DIFF
--- a/schema.graphql
+++ b/schema.graphql
@@ -1,4 +1,5 @@
 scalar Date
+scalar Currency
 
 type Location {
   countryCode: String
@@ -65,6 +66,7 @@ input ShipmentUpdateInput {
   sendingHubId: Int
   receivingHubId: Int
   status: ShipmentStatus
+  pricing: ShipmentPricingInput
 }
 
 type Query {
@@ -104,6 +106,7 @@ input ShipmentCreateInput {
   sendingHubId: Int!
   receivingHubId: Int!
   status: ShipmentStatus!
+  pricing: ShipmentPricingInput
 }
 
 enum ShippingRoute {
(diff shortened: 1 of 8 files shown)


schema.graphql
```diff
@@ -132,13 +135,36 @@ type Shipment {
...
+  singlePallet: MoneyAmount
+  halfPallet: MoneyAmount
+}
+
+input ShipmentPricingInput {
+  singlePallet: MoneyAmountInput
+  halfPallet: MoneyAmountInput
+}
+
+type MoneyAmount {
```
bion, line 156: maybe this should just be called `Money`?

src/tests/helpers/index.ts
```diff
@@ -26,7 +24,10 @@ async function createGroup(
 }
 
 async function createShipment(input: ShipmentCreateInput): Promise<Shipment> {
-  return await Shipment.create(input)
+  return await Shipment.create({
+    ...input,
+    pricing: input.pricing || undefined,
```
deammer, line 29: Question (non-blocking): why is this necessary?

schema.graphql
deammer: I think `MoneyAmount` describes it better

deammer (approved):
💵

src/tests/helpers/index.ts
bion: 😭 because graphql codegen types non-required things as `Thing | null | undefined` and sequelize types non-required things as `Thing | undefined` 😭 😭 😭 😭 😭

bion: See commit message

PR #104, opened 2021-03-21
[wip] Backend work for exporting shipments to google sheets

bion:

PR #119, opened 2021-04-02
Export shipment CSV

bion:
This approach to exporting avoids adding an additional oauth2 flow for authenticating sheets API calls to google (in contrast to #104). The result is code that is simpler to understand.

A design decision here that I am curious for feedback on is the choice to make exports a child of shipments in the graphql schema. I think that's the GQL Way™ as opposed to having an entirely separate list method specifically for exports. The danger for API consumers is that the same query cannot be used for admins and non-admins when viewing the details of a shipment because exports are forbidden for non-admins.


src/server.ts
```diff
@@ -1,25 +1,21 @@
...
-if (process.env.NODE_ENV !== 'production') {
-  dotenv.config()
-}
-
-// Initialize the models
-import './sequelize'
-
-import findOrCreateProfile from './findOrCreateProfile'
+// The autoformatter alphabetizes imports and moves all imports before other code in the file
+import './00_loadEnv'
```
bion, line 7: stupid problems call for stupid solutions

deammer: This is the second time this formatter has caused some annoyance. It better start behaving >:(

schema.graphql
```diff
@@ -141,6 +144,9 @@ type Shipment {
   receivingHub: Group!
   pricing: ShipmentPricing
 
+  # Exports are forbidden to non-users
```
deammer: Nit: non-admins?
```suggestion
  # Exports are forbidden to non-admins
```

deammer (approved):
Nesting `ShipmentExport` under `Shipment` seems fine to me. I agree that it's a little different from how I would have built it outside of GQL but I think it'll make for a nice API on the frontend.

schema.graphql
bion: ah woops, thanks!

PR #127, opened 2021-04-10
Export shipment to CSV button

deammer:
### Changes

- adds a button on the Shipment page to download the shipment's data as a CSV file.
- adds a list of previous shipment exports that can be downloaded
- parse the `createdAt` timestamps as `DateTime` instead of `Date`

### Pixels

Previous versions in the UI:

Screen Shot 2021-04-11 at 10 01 53 AM

Screenshot of a CSV:

Screen Shot 2021-04-10 at 4 54 18 PM

### DIFF
--- a/frontend/.env
+++ b/frontend/.env
@@ -2,7 +2,10 @@
 REACT_APP_VERSION=$npm_package_version
 
 # The URL where the server runs
-REACT_APP_SERVER_URL="http://localhost:3000/graphql"
+REACT_APP_SERVER_URL="http://localhost:3000"
+
+# The URL where the Apollo server runs
+REACT_APP_GRAPHQL_URL="http://localhost:3000/graphql"
 
 # The URL where the client runs
 REACT_APP_CLIENT_URL="http://localhost:8080"
--- a/frontend/src/pages/shipments/DownloadCSVMenu.tsx
+++ b/frontend/src/pages/shipments/DownloadCSVMenu.tsx
@@ -1,0 +1,106 @@
+import { useAuth0 } from '@auth0/auth0-react'
+import format from 'date-fns/format'
+import { FunctionComponent } from 'react'
+import Button from '../../components/Button'
+import PlainModal from '../../components/modal/PlainModal'
+import useModalState from '../../hooks/useModalState'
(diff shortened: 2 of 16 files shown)


frontend/src/pages/shipments/DownloadCSVMenu.tsx
```diff
@@ -0,0 +1,130 @@
...
+        }),
+      )
+      anchorElement.setAttribute('download', fileName)
+      document.body.appendChild(anchorElement)
+      anchorElement.click()
+      document.body.removeChild(anchorElement)
+    } else {
+      window.location.href =
+        'data:application/octet-stream,' + encodeURIComponent(spreadsheet)
+    }
```
bion: Question (blocking): is this necessary? could use the `/shipment-exports/:id` endpoint instead?

bion: If it does need to be done this way, please extract the logic for triggering the download (creating a an anchor el, etc) client-side to a function.

deammer: The tricky part is that I need to send an auth header to that endpoint, which I don't think can be done if I use the `window.location` API?

```ts
window.location.href = "/shipment-exports/123"
```

bion (approved):

frontend/src/pages/shipments/DownloadCSVMenu.tsx
deammer: Okay I refactored this by pointing the client to the endpoint directly, which feels less gross.

frontend/.env
```diff
@@ -2,7 +2,10 @@
 REACT_APP_VERSION=$npm_package_version
 
 # The URL where the server runs
-REACT_APP_SERVER_URL="http://localhost:3000/graphql"
+REACT_APP_SERVER_URL="http://localhost:3000"
+
+# The URL where the Apollo server runs
+REACT_APP_GRAPHQL_URL="http://localhost:3000/graphql"
```
deammer, line 8: This has been updated in Heroku as well!

PR #132, opened 2021-04-12
Create an onboarding flow for new group leaders

### LINKED ISSUE #128: As a new Group Leader, I must create a group
When group leaders create an account, they should be forced to create their own sending group in the UI.

### Pointers

- it would be great to reuse the existing GroupForm component
- Group Leaders should _only_ be able to create a **Sending Group**
- Group Leaders should only be able to create **a single** Sending Group

### Reasoning

Most of the actions performed by a Group Leader will be tied to their group in the backend. If they aren't tied to a group, they will face a bunch of error messages.

deammer:
Resolves #128.

### Changes

- added an onboarding flow for new group leaders
- disabled various UI elements that group leaders shouldn't be allowed to access
- added some restrictions to the Groups API
- added the user's group to their profile information

### Pixels

[🎥 Video showing the new flow from a user perspective](https://www.loom.com/share/f0b29bae40e5407e91d6c0a5f9bf839a)

Home page with onboarding steps:
Screen Shot 2021-04-11 at 9 45 20 PM

Home page after the onboarding is complete:
Screen Shot 2021-04-11 at 9 44 46 PM

### DIFF
--- a/frontend/src/components/TopNavigation.tsx
+++ b/frontend/src/components/TopNavigation.tsx
@@ -1,5 +1,5 @@
 import { useAuth0 } from '@auth0/auth0-react'
-import { FunctionComponent, ReactNode, useContext, useMemo } from 'react'
+import { FunctionComponent, ReactNode, useContext } from 'react'
 import { Link } from 'react-router-dom'
 import ROUTES from '../utils/routes'
 import DistributeAidLogo from './branding/DistributeAidLogo'
@@ -49,12 +49,9 @@ const TopNavigation: FunctionComponent<Props> = ({ hideControls }) => {
   const { user, logout } = useAuth0()
   const { profile } = useContext(UserProfileContext)
 
-  const filteredNavLinks = useMemo(() => {
-    if (!profile?.isAdmin) {
-      return NAV_LINKS.filter((link) => !link.adminOnly)
-    }
-    return NAV_LINKS
-  }, [profile])
+  const filteredNavLinks = NAV_LINKS.filter(
+    (link) => link.adminOnly === profile?.isAdmin,
+  )
 
   return (
     <header className="py-2 bg-navy-800 h-nav sticky top-0">
--- a/frontend/src/components/TopNavigation.tsx
+++ b/frontend/src/components/TopNavigation.tsx
@@ -1,5 +1,5 @@
 import { useAuth0 } from '@auth0/auth0-react'
-import { FunctionComponent, ReactNode } from 'react'
+import { FunctionComponent, ReactNode, useContext, useMemo } from 'react'
 import { Link } from 'react-router-dom'
 import ROUTES from '../utils/routes'
 import DistributeAidLogo from './branding/DistributeAidLogo'
@@ -9,11 +9,13 @@ import TruckIcon from './icons/TruckIcon'
 import UserIcon from './icons/UserIcon'
 import DesktopNavigation from './navigation/DesktopNavigation'
 import MobileNavigation from './navigation/MobileNavigation'
+import { UserProfileContext } from './UserProfileContext'
 
 export interface NavLinkItem {
   path: string
   label: ReactNode
   icon?: ReactNode
+  adminOnly?: boolean
 }
 
 const NAV_LINKS: NavLinkItem[] = [
@@ -26,6 +28,7 @@ const NAV_LINKS: NavLinkItem[] = [
     path: ROUTES.ADMIN_ROOT,
     label: 'Admin',
     icon: <CogIcon className="w-5 h-5 mr-2" />,
+    adminOnly: true,
   },
 ]
 
@@ -44,17 +47,25 @@ interface Props {
  */
 const TopNavigation: FunctionComponent<Props> = ({ hideControls }) => {
   const { user, logout } = useAuth0()
+  const { profile } = useContext(UserProfileContext)
+
+  const filteredNavLinks = useMemo(() => {
+    if (!profile?.isAdmin) {
+      return NAV_LINKS.filter((link) => !link.adminOnly)
+    }
+    return NAV_LINKS
+  }, [profile])
 
   return (
     <header className="py-2 bg-navy-800 h-nav sticky top-0">
       <div className="max-w-5xl px-4 mx-auto h-full flex items-center justify-between">
-        <MobileNavigation navLinks={NAV_LINKS} />
+        <MobileNavigation navLinks={filteredNavLinks} />
         <div className="flex items-center">
           <Link to="/" className="text-white" aria-label="Go to the home page">
             <DistributeAidLogo className="block h-8" />
           </Link>
         </div>
-        <DesktopNavigation navLinks={NAV_LINKS} />
+        <DesktopNavigation navLinks={filteredNavLinks} />
         {!hideControls && user && (
           <div className="flex items-center text-white">
             <DropdownMenu
--- a/frontend/src/components/UserProfileContext.tsx
+++ b/frontend/src/components/UserProfileContext.tsx
@@ -4,6 +4,15 @@ import { createContext, FunctionComponent, useEffect, useState } from 'react'
 export interface UserProfile {
   id: number
   isAdmin: boolean
+  groupId?: number
+}
+
+interface UserProfileData {
+  profile?: UserProfile
+  /**
+   * Call this method to refresh the values stored in the UserProfile
+   */
+  refetch: () => void
 }
 
 const fetchProfile = (token: string) => {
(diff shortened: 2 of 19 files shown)


frontend/src/components/TopNavigation.tsx
```diff
@@ -44,17 +47,25 @@ interface Props {
...
 const TopNavigation: FunctionComponent<Props> = ({ hideControls }) => {
   const { user, logout } = useAuth0()
+  const { profile } = useContext(UserProfileContext)
+
+  const filteredNavLinks = useMemo(() => {
+    if (!profile?.isAdmin) {
+      return NAV_LINKS.filter((link) => !link.adminOnly)
+    }
+    return NAV_LINKS
+  }, [profile])
```
deammer: Some nav links are only accessible to admins.

frontend/src/components/UserProfileContext.tsx
```diff
@@ -39,7 +52,7 @@ const UserProfileProvider: FunctionComponent = ({ children }) => {
   }, [isAuthenticated, getAccessTokenSilently, profileIsLoaded])
 
   return (
-    <UserProfileContext.Provider value={profile}>
+    <UserProfileContext.Provider value={{ profile, refetch }}>
```
deammer, line 55: I had to pass down the `refetch` method so you'll see a lot of diffs like:

```diff
- const profile = useContext(UserProfileContext)
+ const { profile } = useContext(UserProfileContext)
```

frontend/src/pages/groups/GroupCreatePage.tsx
```diff
@@ -10,27 +11,35 @@ import { groupViewRoute } from '../../utils/routes'
...
+      const { data } = await addGroup({
+        variables: { input },
+        // Fetch the updated list of groups
+        refetchQueries: [{ query: AllGroupsDocument }],
       })
-      .catch(console.error)
+
+      // If the user is a group leader, refresh their credentials so that the UI
+      // is aware that they created their own group and finished their
+      // onboarding
```
deammer: Does this make sense/is it clear enough?

frontend/src/pages/home/GroupLeaderHomePage.tsx
```diff
@@ -0,0 +1,73 @@
+import { FunctionComponent, useContext } from 'react'
+import ButtonLink from '../../components/ButtonLink'
+import CheckIcon from '../../components/icons/CheckIcon'
+import CogIcon from '../../components/icons/CogIcon'
+import TruckIcon from '../../components/icons/TruckIcon'
+import InternalLink from '../../components/InternalLink'
+import { UserProfileContext } from '../../components/UserProfileContext'
+import ROUTES, { groupViewRoute } from '../../utils/routes'
+
+const GroupLeaderHomePage: FunctionComponent = () => {
```
deammer, line 10: Group leaders have their own home page! See screenshots in the PR description.

src/findOrCreateProfile.ts
```diff
@@ -1,20 +1,31 @@
...
-    sendProfile(response, auth.userAccount, auth.isAdmin)
+    let groupForUser: Maybe<Group>
+    if (!auth.isAdmin) {
+      groupForUser = await Group.findOne({
+        where: { captainId: auth.userAccount.id },
+        attributes: ['id'],
+      })
+    }
+
+    sendProfile(response, auth.userAccount, auth.isAdmin, groupForUser?.id)
```
deammer: I'd love some feedback on this because there's probably a better way to do this. The goal is to send down the user's `groupId` (if they have one) so that the UI can figure out whether they need onboarding or not.

frontend/src/components/TopNavigation.tsx
```diff
@@ -44,17 +47,25 @@ interface Props {
  */
 const TopNavigation: FunctionComponent<Props> = ({ hideControls }) => {
   const { user, logout } = useAuth0()
+  const { profile } = useContext(UserProfileContext)
+
+  const filteredNavLinks = useMemo(() => {
```
bion: Suggestion (non-blocking): this isn't a strong opinion, but imo the additional code complexity and hit to readability isn't worth the performance improvement of memoizing things like this where N is likely to stay very small. In general I'd recommend only memoizing things that have a measurable performance improvement.

frontend/src/pages/groups/GroupCreatePage.tsx
bion: I'm not clear on why/how `refetch` does this. The comment beginning with "If" leads me to expect a conditional.

Suggestion: add a sentence that begins with "refetch here will..."

frontend/src/pages/groups/GroupForm.tsx
```diff
@@ -30,40 +32,53 @@ interface Props {
...
     formState: { errors },
-  } = useForm()
+  } = useForm<GroupCreateInput>()
 
   useEffect(
     function resetFormValues() {
       if (props.defaultValues) {
         // Update the values of the fields
-        reset(props.defaultValues.group)
+        reset(_omit(props.defaultValues.group, ['id', '__typename']))
```
bion, line 48: Suggestion (non-blocking): could extract this to a shared function called `omitGraphqlTypeName` or something

frontend/src/pages/groups/GroupForm.tsx
```diff
@@ -30,40 +32,53 @@ interface Props {
...
       }
     },
     [props.defaultValues, reset],
   )
 
+  const submitForm = handleSubmit((input) => {
+    // Non-admins can only create sending groups
+    if (!profile?.isAdmin) {
```
bion, line 56: Question (non-blocking): what trade-offs would there be to ensure that there is a profile available here and where other code that is uncertain about the presence of a profile runs? Just so we're not constantly using `?.`.  My concern is that we'll get accustomed to using `?.` so much we'll miss that someone actually missing a profile will get an experience they shouldn't.

frontend/src/pages/home/GroupLeaderHomePage.tsx
bion: awesome!

src/findOrCreateProfile.ts
bion: The incongruity here is that in the backend there's a one-to-many association between user accounts and groups. I don't know if we need/want to support a person being a member of multiple groups. If we do it'd likely complicate the UI in addition to code you've added in this PR.

The code as-written is mostly fine so long as we don't somehow end up with a user with multiple groups in the DB, which is possible at least as far as the backend is concerned.

I'd be ok with more or less leaving it as is while we gather feedback and pay the tech debt in the future if need be to fully support one-to-many rather than just one-to-one. Short term, if we *don't* need to support one-to-many we should add validation checks in `src/resolvers/group.ts` to ensure it doesn't happen since right now I think the admin could update multiple groups to have the same captain..

src/models/user_account.ts
```diff
@@ -36,10 +36,11 @@ export default class UserAccount extends Model<
   @Column
   public readonly updatedAt!: Date
 
-  public asProfile(isAdmin = false): UserProfile {
+  public asProfile(isAdmin = false, groupId?: number): UserProfile {
```
bion: Suggestion (blocking): swap these so the param with a default is second so the caller can omit it if they're ok with the default.

src/resolvers/group.ts
```diff
@@ -41,6 +42,11 @@ const addGroup: MutationResolvers['addGroup'] = async (
     })
   }
 
+  // Non-admins should only be allowed to create sending groups
+  if (input.groupType !== GroupType.SendingGroup && !context.auth.isAdmin) {
+    throw new ForbiddenError(`Non-admins can only create sending groups`)
+  }
```
bion, line 48: 👍🏻

src/resolvers/group.ts
```diff
@@ -73,7 +79,11 @@ const updateGroup: MutationResolvers['updateGroup'] = async (
     throw new ForbiddenError('Not permitted to update group')
   }
 
-  if (input.groupType && !context.auth.isAdmin) {
+  if (
+    input.groupType &&
+    input.groupType !== group.groupType &&
+    !context.auth.isAdmin
```
bion, line 85: 👍🏻

bion (approved):

src/findOrCreateProfile.ts
deammer: I appreciate your thought process. I agree with shipping this as-is and figuring out the one-to-many vs one-to-one relationship in the future. I'll create an issue and add a comment in the code.

frontend/src/pages/groups/GroupForm.tsx
deammer: That's a great question and I'd like to do a little spike on it, so I opened #134.

frontend/src/pages/groups/GroupForm.tsx
deammer: This is the only place where we're doing that so I'll leave it as is, but I'll create a function if we end up repeating that same code in the future 👍

PR #135, opened 2021-04-14
Updated the Auth0 client ID and added docs

deammer:
Changed the Auth0 client ID to match the new app we created through https://github.com/distributeaid/auth0/commit/7f12faa07b21f0dbe9eddb1b9fe5fe417fb3deee


deammer:
Merged without review since it's a blocker and it's low-risk.

PR #143, opened 2021-05-03
Prevent users from creating shipments in the past

### LINKED ISSUE #140: [bug] It is possible to create shipments for past dates
The UI and the API should not allow users to create shipments with past dates.

deammer:
Resolves #140 and prevents the creation of causal loops.

### DIFF
--- a/frontend/src/pages/shipments/ShipmentForm.tsx
+++ b/frontend/src/pages/shipments/ShipmentForm.tsx
@@ -1,8 +1,8 @@
+import _range from 'lodash/range'
 import { FunctionComponent, ReactNode, useEffect, useState } from 'react'
 import { useForm } from 'react-hook-form'
 import Button from '../../components/Button'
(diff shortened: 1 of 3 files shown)


frontend/src/pages/shipments/ShipmentForm.tsx
```diff
@@ -125,22 +142,28 @@ const ShipmentForm: FunctionComponent<Props> = (props) => {
...
           castAsNumber
           register={register}
+          registerOptions={{
+            validate: validateShipmentDate,
+          }}
           required
           errors={errors}
         />
-        <TextField
+        <SelectField
```
deammer, line 157: The year is now a dropdown, which is a little quicker to fill out and easier to validate.

Screen Shot 2021-05-03 at 4 29 50 PM

bion (approved):

PR #145, opened 2021-05-13
Improvements to the line item form

deammer:
### Changes

Thanks to Dom and Taylor's review, we decided to make a bunch of tiny changes to the line item form.

- renamed `containerWeightGrams` to `containerWeightKilos` because we don't ever expect to deal in grams
- added a disclaimer about the maximum dimensions of pallets
- pre-fill some of the line item dimensions under certain circumstances
- renamed some form fields based on the values of other fields
- show the description of line items in the sidebar


db/migrations/20210513002725-update-container-weight-unit.js
```diff
@@ -0,0 +1,19 @@
...
+  },
+
+  down: async (queryInterface, Sequelize) => {
+    return await queryInterface.renameColumn(
+      'LineItems',
+      'containerWeightKilos',
+      'containerWeightGrams',
+    )
+  },
+}
```
bion: Question (blocking): will we ever want more precision than 1 kilo? e.g. 1.5 kilos? If so we should leave this as is. The column is an integer type to avoid floating point semantics.

deammer: Yeah I was on the edge about that 🤔 I'll try an alternative.

deammer: @bion I reverted the backend change and handled the weight conversion on the frontend.

bion (approved):

PR #149, opened 2021-05-28
Fixed the Auth0 cookie expiration problem

### LINKED ISSUE #139: [bug] Refreshing the page shows the login screen
> Refresh takes me back to the "login" page, but when I click login it takes me to the admin dashboard.  Can we just do an auth check to skip the extra step if I'm already logged in?

This happened in the staging environment: https://shipment-tracker-dev.herokuapp.com/

deammer:
Fixes #139 

Because the Auth0 cookie expires after a day (or something), the website often showed the login page even though it was possible to silently re-auth the user.

The trick was to silently attempt to authenticate the user in the background.


bion (approved):
Looks good, maybe add a comment mentioning re-auth?

Issue #47, opened 2021-03-01
Add API input validation


andyjac:
Add input validation (see #152, based on [this suggested pattern](https://github.com/distributeaid/shipment-tracker/pull/43#discussion_r584276670)) to these resolvers:
- [x] [addGroup](https://github.com/distributeaid/shipment-tracker/blob/413c3e76c05c4ea70f3c5ba4d995c67f7f06edf1/src/resolvers/group.ts#L69) #152 
- [x] [updateGroup](https://github.com/distributeaid/shipment-tracker/blob/413c3e76c05c4ea70f3c5ba4d995c67f7f06edf1/src/resolvers/group.ts#L104) #162
- [x] [group](https://github.com/distributeaid/shipment-tracker/blob/413c3e76c05c4ea70f3c5ba4d995c67f7f06edf1/src/resolvers/group.ts#L19) #162
- [x] [addOffer](https://github.com/distributeaid/shipment-tracker/blob/dd4252dac9c119724d08f813c3a07a1ffc05f616/src/resolvers/offer.ts#L21) #163
- [x] [updateOffer](https://github.com/distributeaid/shipment-tracker/blob/dd4252dac9c119724d08f813c3a07a1ffc05f616/src/resolvers/offer.ts#L76) #163
- [x] [offer](https://github.com/distributeaid/shipment-tracker/blob/dd4252dac9c119724d08f813c3a07a1ffc05f616/src/resolvers/offer.ts#L112) #163
- [x] [listOffers](https://github.com/distributeaid/shipment-tracker/blob/dd4252dac9c119724d08f813c3a07a1ffc05f616/src/resolvers/offer.ts#L126) #163
- [x] [lineItem](https://github.com/distributeaid/shipment-tracker/blob/8d1643171c944fe80229c60b45b25dd5cdd372ff/src/resolvers/line_items.ts#L27) #168
- [x] [addLineItem](https://github.com/distributeaid/shipment-tracker/blob/8d1643171c944fe80229c60b45b25dd5cdd372ff/src/resolvers/line_items.ts#L44) #168
- [x] [updateLineItem](https://github.com/distributeaid/shipment-tracker/blob/8d1643171c944fe80229c60b45b25dd5cdd372ff/src/resolvers/line_items.ts#L75) #168
[...]

PR #160, opened 2021-06-13
feat(input-validation): refactor and add country codes

coderbyheart:
This extracts the types used for addGroup input validation
into reusable types and adds [country codes](https://www.iso.org/obp/ui/#search) to tighten
input validation for this input to validate against
a known list of country codes.

See #47


coderbyheart:
Fun fact, `UK` is not a valid country code for Englang/United Kingdom/Great Britain according to [ISO 3166](https://en.wikipedia.org/wiki/ISO_3166-1#Coding).

@jtfairbank is it critical to keep the `UK` on the database level? If yes, I can add an alias entry.

deammer (approved):
Looks good!

deammer:
I'd rather we don't add an alias system just to support `UK`. I don't think the added complexity is worth it.

PR #164, opened 2021-06-13
feat(input-validation): add utility to make errors human readable

coderbyheart:
This adds errorsToProblemDetails() which turns schema validation errors
into better human readable Problem Details for HTTP APIs.
See https://datatracker.ietf.org/doc/html/rfc7807

Depends on #163


deammer (approved):
This is awesome 🙏

PR #197, opened 2021-06-24
Added 3 shipment statuses

deammer:
### Changes

Added some shipment statuses:

```diff
 enum ShipmentStatus {
+  DRAFT
   ANNOUNCED
   OPEN
+  AID_MATCHING
   STAGING
   IN_PROGRESS
   COMPLETE
   ABANDONED
+  ARCHIVED
 }
```

### Screenshot

I formatted the statuses so things look a little nicer and less YELLY

Screen Shot 2021-06-23 at 5 18 09 PM

### DIFF
--- a/frontend/src/components/TopNavigation.tsx
+++ b/frontend/src/components/TopNavigation.tsx
@@ -57,7 +57,7 @@ const TopNavigation: FunctionComponent<Props> = ({ hideControls }) => {
   )
 
   return (
-    <header className="py-2 bg-navy-800 h-nav sticky top-0">
+    <header className="py-2 bg-navy-800 h-nav sticky top-0 z-20">
       <div className="max-w-5xl px-4 mx-auto h-full flex items-center justify-between">
         <MobileNavigation navLinks={filteredNavLinks} />
         <div className="flex items-center">
--- a/frontend/src/data/constants.ts
+++ b/frontend/src/data/constants.ts
@@ -46,13 +46,23 @@ export const OFFER_STATUS_OPTIONS = enumValues(OfferStatus).map((routeKey) => ({
   value: routeKey,
 }))
 
+/**
+ * Order of shipment statuses:
+ * draft → announced → open to offers → aid matching → staging at a hub
+ * → in transit → complete
+ *
+ * A shipment can also be 'abandoned' or 'archived' at any point.
+ */
 export const SHIPMENT_STATUS_OPTIONS = [
-  { label: 'Abandoned', value: ShipmentStatus.Abandoned },
+  { label: 'Draft', value: ShipmentStatus.Draft },
   { label: 'Announced', value: ShipmentStatus.Announced },
+  { label: 'Open to offers', value: ShipmentStatus.Open },
+  { label: 'Aid matching', value: ShipmentStatus.AidMatching },
+  { label: 'Staging at a hub', value: ShipmentStatus.Staging },
+  { label: 'In transit', value: ShipmentStatus.InProgress },
   { label: 'Complete', value: ShipmentStatus.Complete },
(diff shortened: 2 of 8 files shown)


frontend/src/components/TopNavigation.tsx
```diff
@@ -57,7 +57,7 @@ const TopNavigation: FunctionComponent<Props> = ({ hideControls }) => {
   )
 
   return (
-    <header className="py-2 bg-navy-800 h-nav sticky top-0">
+    <header className="py-2 bg-navy-800 h-nav sticky top-0 z-20">
```
deammer, line 60: Unrelated UI bug where some elements with `position: absolute` would display on top of the sticky nav because they appear later in the HTML.

frontend/src/data/constants.ts
```diff
@@ -46,13 +46,23 @@ export const OFFER_STATUS_OPTIONS = enumValues(OfferStatus).map((routeKey) => ({
   value: routeKey,
 }))
 
+/**
+ * Order of shipment statuses:
+ * draft → announced → open to offers → aid matching → staging at a hub
+ * → in transit → complete
```
deammer, line 52: This is the result of a long discussion with Dom and Elliot 😄

frontend/src/pages/shipments/ShipmentForm.tsx
```diff
@@ -38,15 +41,6 @@ interface Props {
...
 
-const STATUS_OPTIONS = [
-  { label: 'Announced', value: ShipmentStatus.Announced },
-  { label: 'In progress', value: ShipmentStatus.InProgress },
-  { label: 'Open', value: ShipmentStatus.Open },
-  { label: 'In staging', value: ShipmentStatus.Staging },
-  { label: 'Complete', value: ShipmentStatus.Complete },
-  { label: 'Abandoned', value: ShipmentStatus.Abandoned },
-]
-
```
deammer, line 49: This was duplicated in the constants.ts file

frontend/src/pages/shipments/ShipmentForm.tsx
```diff
@@ -132,15 +128,16 @@ const ShipmentForm: FunctionComponent<Props> = (props) => {
...
-      <SelectField
-        defaultValue={ShipmentStatus.Announced}
-        options={STATUS_OPTIONS}
-        label="Status"
-        name="status"
-        register={register}
-        required
-        errors={errors}
-      />
+      {isExistingShipment && (
```
deammer, line 131: Users can only set a shipment's status when editing it, not creating it.

frontend/src/data/constants.ts
coderbyheart: **thought:** we could express this as a state machine (e.g. using [xstate](https://xstate.js.org/) and validate state transitions (both in the UI and the backend) using that. This is only useful if there are actually restrictions on how the state can be change.

frontend/src/utils/format.ts
```diff
@@ -124,6 +120,13 @@ export function formatShippingRouteName(shippingRoute: ShippingRoute) {
   return matchingRoute?.label || 'Unknown route'
 }
 
+export function formatShipmentStatus(shipmentStatus: ShipmentStatus) {
+  const matchingStatus = SHIPMENT_STATUS_OPTIONS.find(
+    (option) => option.value === shipmentStatus,
+  )
+  return matchingStatus?.label || shipmentStatus
```
coderbyheart, line 127:
```suggestion
  return matchingStatus?.label ?? shipmentStatus
```
*nit:* use nullish coalescing operator, because it only handles cases of `null` or `undefined`, and this is the intention here.

coderbyheart (approved):

PR #212, opened 2021-06-27
Fix yarn lock file

deammer:
Fixes this deployment issue:
```
-----> Build failed

 !     Outdated Yarn lockfile

       Your application contains a Yarn lockfile (yarn.lock) which does not

       match the dependencies in package.json. This can happen if you use npm

       to install or update a dependency instead of Yarn.
```

Resolves #181 
Resolves #184 
Resolves #191 
Resolves #192 
Resolves #193 
Resolves #194 
Resolves #195 
Resolves #198 
Resolves #199 
Resolves #207
Resolves #208 
Resolves #210 
Resolves #211


coderbyheart:
I wonder why this was not caught before?
After all CI runs yarn build: https://github.com/distributeaid/shipment-tracker/blob/d9fd09ba418a515b92d6da4b5ec1978541cc2b2b/.github/workflows/ci.yml#L53

coderbyheart (approved):

PR #213, opened 2021-06-28
Allow captains to create shipments

### LINKED ISSUE #131: As a Group Leader, I can only create offers for an OPEN shipment
Group leaders should only be able to create offers for a shipment in the `OPEN` status. This restriction should be enforced on the backend and frontend.

For example: if a shipment has the `DRAFT` status, a group captain should _not_ be allowed to create offers for that shipment.

coderbyheart:
Closes #131


coderbyheart:
The point of #131 is to only allow the **creation of OFFERS** for **OPEN shipments**.

PR #168, opened 2021-06-20
feat(input-validation): add validation to line-item

coderbyheart:
See #47

PR #229, opened 2021-07-05
Standardize the removal of null and empty fields on the frontend

### LINKED ISSUE #216: Updating a Line Items's receiving group doesn't work

deammer:
This refactor stems from [this discussion](https://github.com/distributeaid/shipment-tracker/pull/168#discussion_r657333035). We want to remove empty/null fields before sending them to the backend. This PR adds a utility method to do just that.

This PR also resolves #216 and adds support for unit tests on the frontend through a GitHub action.

### DIFF
--- a/frontend/babel.config.js
+++ b/frontend/babel.config.js
@@ -1,0 +1,6 @@
+module.exports = {
+  presets: [
+    ['@babel/preset-env', { targets: { node: 'current' } }],
+    '@babel/preset-typescript',
+  ],
+}
--- a/frontend/src/utils/data.ts
+++ b/frontend/src/utils/data.ts
@@ -1,0 +1,24 @@
+/**
+ * Strips empty, undefined, and null fields from an object. This method is
+ * intentionally NOT recursive.
+ *
+ * @example setEmptyFieldsToUndefined({
+ *   a: "",
+ *   b: null,
+ *   c: 123
+ * })
+ * // returns { c: 123 }
+ */
+export function setEmptyFieldsToUndefined<T>(data: T) {
+  if (typeof data === 'object') {
+    const objectWithoutNulls = { ...data } as any
+    Object.keys(data).forEach((key) => {
+      if (objectWithoutNulls[key] == null || objectWithoutNulls[key] === '') {
+        delete objectWithoutNulls[key]
+      }
+    })
+    return objectWithoutNulls as T
+  }
+
+  return data
+}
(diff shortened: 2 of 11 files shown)


src/resolvers/line_items.ts
```diff
@@ -256,22 +252,12 @@ async function getUpdateAttributes(
...
-  updateAttributes: Partial<LineItemAttributes>,
   attr: 'acceptedReceivingGroupId' | 'proposedReceivingGroupId',
 ) {
-  if (
-    !('acceptedReceivingGroupId' in input) ||
-    input[attr] === lineItem[attr]
-  ) {
-    return
-  }
-
```
deammer, line 271: I removed this because if the value stayed the same, it would get deleted in the database, which isn't what we wanted.

src/resolvers/line_items.ts
```diff
@@ -388,38 +374,36 @@ const moveLineItem: MutationResolvers['moveLineItem'] = async (
 }
 
 // Line item custom resolvers
-const proposedReceivingGroup: LineItemResolvers['proposedReceivingGroup'] = async (
-  parent,
-) => {
-  if (!parent.proposedReceivingGroupId) {
-    return null
-  }
+const proposedReceivingGroup: LineItemResolvers['proposedReceivingGroup'] =
```
deammer, line 377: I didn't change anything in this resolver, so our code formatters are probably out of sync.

src/resolvers/line_items.ts
deammer: This is the fix for #216.

frontend/src/utils/data.ts
```diff
@@ -0,0 +1,24 @@
...
+ * Strips empty, undefined, and null fields from an object. This method is
+ * intentionally NOT recursive.
+ *
+ * @example setEmptyFieldsToUndefined({
+ *   a: "",
+ *   b: null,
+ *   c: 123
+ * })
+ * // returns { c: 123 }
+ */
```
coderbyheart, line 11: **suggestion:** turn this into a unit test. It will serve as an example AND verify that `setEmptyFieldsToUndefined` does what it's supposed to be.

coderbyheart (approved):

frontend/babel.config.js
```diff
@@ -0,0 +1,6 @@
+module.exports = {
+  presets: [
+    ['@babel/preset-env', { targets: { node: 'current' } }],
+    '@babel/preset-typescript',
+  ],
+}
```
deammer, line 6: Jest requires Babel to parse TypeScript code.

frontend/src/utils/data.ts
deammer: Good idea, and done.

PR #251, opened 2021-07-13
Fix the logic of the offer creation page

### LINKED ISSUE #230: The Create Offer form is broken
The CreateOfferForm breaks because the contact info isn't filled out automatically.

While we're there, it might also be nice to build a separate form for admins since the functionality is so different. There are a few `TODO`s in the `CreateOfferForm.tsx` file.

deammer:
This PR addresses 2 problem:
- the fact that some users have access to multiple groups (ie admins) and should thus be allowed to create offers on any of those groups
- bug #230 where the contact info doesn't fill out automatically and prevents users from creating offers

Fixes #230.

### DIFF
--- a/frontend/src/pages/groups/allGroupsQuery.graphql
+++ b/frontend/src/pages/groups/allGroupsQuery.graphql
@@ -1,5 +1,5 @@
-query AllGroups {
-  listGroups {
+query AllGroups($captainId: Int, $groupType: [GroupType!]) {
+  listGroups(captainId: $captainId, groupType: $groupType) {
     id
     name
     groupType
--- a/frontend/src/pages/offers/CreateOfferForm.tsx
+++ b/frontend/src/pages/offers/CreateOfferForm.tsx
@@ -1,14 +1,16 @@
-import { FunctionComponent, useContext, useMemo } from 'react'
+import _pick from 'lodash/pick'
+import { FunctionComponent, useContext, useEffect } from 'react'
 import { useForm } from 'react-hook-form'
 import Button from '../../components/Button'
 import ReadOnlyField from '../../components/forms/ReadOnlyField'
+import SelectField from '../../components/forms/SelectField'
 import TextField from '../../components/forms/TextField'
 import Spinner from '../../components/Spinner'
 import { UserProfileContext } from '../../components/UserProfileContext'
 import {
   GroupType,
   OfferCreateInput,
-  useAllGroupsQuery,
+  useAllGroupsLazyQuery,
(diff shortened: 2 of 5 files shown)


frontend/src/pages/groups/allGroupsQuery.graphql
```diff
@@ -1,5 +1,5 @@
-query AllGroups {
-  listGroups {
+query AllGroups($captainId: Int, $groupType: [GroupType!]) {
```
deammer, line 1: Markus built the functionality in the resolver, so this is all it takes to add filtering on the client.

frontend/src/pages/offers/CreateOfferForm.tsx
```diff
@@ -29,88 +31,126 @@ interface Props {
...
- * If the user is not an admin, they should automatically use their own group
- */
-
 /**
  * This form allows users to create a new offer. It is very specific to offer
  * creation, and therefore difficult to reuse for editing offers.
+ *
+ * There is a situation that makes things complex:
+ * - some users (usually admins) have the ability to create offers for any group
+ * - other users only have access to a single sending group
```
deammer, line 40: Yeah this was fun. Happy to answer any questions!

frontend/src/pages/offers/CreateOfferForm.tsx
```diff
@@ -29,88 +31,126 @@ interface Props {
...
+      </div>
+    )
   }
 
   return (
     <form onSubmit={handleSubmit(onSubmitForm)} className="space-y-6">
-      <ReadOnlyField label="Sending group">
-        {groupForUser?.name || 'No group'}
-      </ReadOnlyField>
+      {groups.listGroups.length > 1 ? (
```
deammer, line 137: If the user has access to more than 1 group, we display a dropdown. Otherwise, we select the group for them.

src/resolvers/offer.ts
```diff
@@ -56,7 +56,10 @@ const addOffer: MutationResolvers['addOffer'] = async (
   })
 
   const sendingGroup = await sendingGroupPromise
-  if (sendingGroup?.captainId !== context.auth.userAccount.id) {
+  if (
+    sendingGroup?.captainId !== context.auth.userAccount.id &&
+    !context.auth.isAdmin
```
deammer, line 61: This allows admins to create offers for any group.

coderbyheart, line 61: Needs a test.

coderbyheart:
I've added a test.

PR #252, opened 2021-07-17
Filter the list of shipments for group captains

### LINKED ISSUE #130: As a Group Leader, I can only see a subset of all shipments
Group leaders should only see shipments with a status that's one of `ANNOUNCED`, `OPEN`, or `STAGING`.

### Notes

- admins should remain unaffected
- this depends on the work in #129

deammer:
Limits which shipments group captains can see in the list. Resolves #130.

Note: I originally assumed it'd be easier to do this on the client, but it was much cleaner to do it on the backend 😅

### DIFF
--- a/src/resolvers/shipment.ts
+++ b/src/resolvers/shipment.ts
@@ -33,12 +33,41 @@ const listShipmentsInput = Type.Object(
   { additionalProperties: false },
 )
 
+const SHIPMENT_STATUSES_ALLOWED_FOR_CAPTAINS = [
+  ShipmentStatus.Announced,
+  ShipmentStatus.Open,
+  ShipmentStatus.Staging,
+]
+
 const validateListShipmentsInput = validateWithJSONSchema(listShipmentsInput)
 
-const listShipments: QueryResolvers['listShipments'] = async (_, input) => {
+const listShipments: QueryResolvers['listShipments'] = async (
+  _,
+  input,
+  context,
+) => {
   const valid = validateListShipmentsInput(input)
   if ('errors' in valid) {
     throw new UserInputError('List shipments input invalid', valid.errors)
+  }
+
+  if (!context.auth.isAdmin) {
+    // Group captains can only view a subset of shipments by status
+    if (valid.value.status == null) {
+      return Shipment.findAll({
+        where: {
+          status: SHIPMENT_STATUSES_ALLOWED_FOR_CAPTAINS,
+        },
+      })
+    } else {
+      return Shipment.findAll({
+        where: {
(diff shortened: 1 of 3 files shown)


src/resolvers/shipment.ts
```diff
@@ -33,14 +33,43 @@ const listShipmentsInput = Type.Object(
...
+    if (valid.value.status == null) {
+      return Shipment.findAll({
+        where: {
+          status: SHIPMENT_STATUSES_ALLOWED_FOR_CAPTAINS,
+        },
+      })
+    } else {
+      return Shipment.findAll({
+        where: {
+          status: valid.value.status.filter((status) =>
```
coderbyheart: If captain requests to see a shipment with a status they are not supposed to see, that request should be rejected with a forbidden error.

coderbyheart (changes requested):

src/resolvers/shipment.ts
```diff
@@ -33,14 +33,43 @@ const listShipmentsInput = Type.Object(
...
+  input,
+  context,
+) => {
   const valid = validateListShipmentsInput(input)
   if ('errors' in valid) {
     throw new UserInputError('List shipments input invalid', valid.errors)
   }
 
+  if (!context.auth.isAdmin) {
+    // Group captains can only view a subset of shipments by status
```
coderbyheart: It is not established that the calling user is a group captain.

src/resolvers/shipment.ts
coderbyheart: If only certain status are allowed for listing, this needs to be applied to fetching individual shiptments.

coderbyheart:
@deammer I've added tests to the backend logic and changed the API to throw a forbidden error, if a non-admin requests a status they are not supposed to see.

One thing I noticed is that you write about "captains", but at no point was the captain role established. So I changed the wording, to match the actual logic. The limitation now distinguishes between admins and users.

coderbyheart (approved):

deammer:
@coderbyheart thank you, I really appreciate your help!

coderbyheart:
@deammer next time when you are about to merge a PR, please check if you are up to date with saga. The merge resulted in broken code.

deammer:
That's weird because it's impossible to merge through the UI when a branch is out of date with saga 🤔 I wonder if maybe the page needed to be refreshed or something.

Screen Shot 2021-07-19 at 9 25 34 AM

coderbyheart:
I think this only applies if there is a conflict. In case of 552bf7eb1f9eff2cf79cf0fcc887093a31ff29af there was no conflict, the changes could be applied.

PR #260, opened 2021-07-19
Validate pallet contents and display a summary

### LINKED ISSUE #258: Refactor the UI for line item creation
The current UI asks the user to select a container type when creating a line item. One of the options is "pallet", which doesn't make sense because line items are already nested under a pallet.

Additionally, pallet contents should follow the following restrictions:
- maximum of 36 boxes per pallet
- maximum of 1 bulk bag per pallet
- if a pallet contains a bulk bag, it can contain a maximum of 18 boxes

deammer:
Resolves #258.

### Changes

Improved the UX around the pallet layout. I hope the layout is less confusing.

|With boxes|Without boxes|
|---|---|
|Screen Shot 2021-07-21 at 10 57 18 AM|Screen Shot 2021-07-21 at 10 57 14 AM|

Added a pallet summary with a validator:

|Valid contents|Invalid contents|
|---|---|
|Screen Shot 2021-07-21 at 10 57 24 AM|Screen Shot 2021-07-21 at 10 58 09 AM|

Other changes:
- refactored some markup into components such as the `PalletsEditorSidebar`
- added unit tests for the `validatePalletContents` utility

### DIFF
--- a/.github/workflows/frontend-unit-tests.yml
+++ b/.github/workflows/frontend-unit-tests.yml
@@ -2,10 +2,6 @@ name: Frontend unit tests
 
 on: push
 
-defaults:
-  run:
-    working-directory: frontend
-
 jobs:
   build:
     runs-on: ubuntu-latest
@@ -17,6 +13,12 @@ jobs:
     steps:
       - uses: actions/checkout@v2
       - name: Install modules
-        run: yarn
-      - name: Run tests
+        run: yarn --frozen-lockfile
+      - name: Run codegen
+        run: yarn codegen
+      - name: Install frontend modules
+        working-directory: frontend
+        run: yarn --frozen-lockfile
+      - name: Run unit tests
+        working-directory: frontend
         run: yarn test
--- a/frontend/src/pages/groups/allGroupsQuery.graphql
+++ b/frontend/src/pages/groups/allGroupsQuery.graphql
@@ -18,8 +18,8 @@ query AllGroups($captainId: Int, $groupType: [GroupType!]) {
   }
 }
 
-query AllGroupsMinimal {
-  listGroups {
+query AllGroupsMinimal($captainId: Int, $groupType: [GroupType!]) {
+  listGroups(captainId: $captainId, groupType: $groupType) {
     id
     name
     groupType
--- a/frontend/src/pages/offers/LineItemForm.tsx
+++ b/frontend/src/pages/offers/LineItemForm.tsx
@@ -10,7 +10,6 @@ import Spinner from '../../components/Spinner'
 import {
   DANGEROUS_GOODS_LIST,
   LINE_ITEM_CATEGORY_OPTIONS,
-  LINE_ITEM_CONTAINER_OPTIONS,
   PALLET_CONFIGS,
 } from '../../data/constants'
 import {
@@ -26,7 +25,6 @@ import {
 } from '../../types/api-types'
 import { setEmptyFieldsToUndefined } from '../../utils/data'
 import {
-  getContainerCountLabel,
   gramsToKilos,
   groupToSelectOption,
   kilosToGrams,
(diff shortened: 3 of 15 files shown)


frontend/src/utils/data.ts
```diff
@@ -22,3 +25,69 @@ export function setEmptyFieldsToUndefined<T>(data: T) {
 
   return data
 }
+
+export function validatePalletContents(pallet: PalletQuery['pallet']) {
```
coderbyheart: **idea:** would be cool to express these business rules in a way so they can be used in the backend as well.

deammer: @coderbyheart I totally agree. One hurdle is that the frontend can't import files from the top-level directory, so we'd have to figure that out.

That's the reason why the codegen system generates similar files in two directories.

coderbyheart: If it were to be put on the top-level, it could be linked as a module via the package.json. Also TypeScript has a feature for that: https://www.typescriptlang.org/tsconfig#paths

.github/workflows/frontend-unit-tests.yml
```diff
@@ -17,6 +13,12 @@ jobs:
     steps:
       - uses: actions/checkout@v2
       - name: Install modules
-        run: yarn
-      - name: Run tests
+        run: yarn --frozen-lockfile
+      - name: Run codegen
+        run: yarn codegen
```
deammer, line 18: We need to run `codegen` to run unit tests because they rely on auto-generated imports. See utils.test.ts in this PR.

frontend/src/pages/groups/allGroupsQuery.graphql
```diff
@@ -18,8 +18,8 @@ query AllGroups($captainId: Int, $groupType: [GroupType!]) {
   }
 }
 
-query AllGroupsMinimal {
-  listGroups {
+query AllGroupsMinimal($captainId: Int, $groupType: [GroupType!]) {
+  listGroups(captainId: $captainId, groupType: $groupType) {
```
deammer, line 22: This adds the ability to optionally filter the list of groups.

frontend/src/pages/offers/LineItemForm.tsx
```diff
@@ -63,15 +61,14 @@ const LineItemForm: FunctionComponent<Props> = ({
...
+
   useEffect(
     function organizeGroups() {
       if (groups && groups.listGroups) {
-        setReceivingGroups(
-          groups.listGroups
-            .filter((group) => group.groupType === GroupType.ReceivingGroup)
-            .map(groupToSelectOption),
-        )
+        setReceivingGroups(groups.listGroups.map(groupToSelectOption))
```
deammer: The functionality remains the same, but we filter the group type on the backend instead of on the client.

frontend/src/pages/offers/LineItemForm.tsx
```diff
@@ -282,8 +285,8 @@ const LineItemForm: FunctionComponent<Props> = ({
       ) : (
         <PlaceholderField className="mb-6 max-w-md" />
       )}
-      <fieldset className="space-y-4">
-        <legend className="font-semibold text-gray-700 ">Contents</legend>
+      <fieldset className="space-y-4 mb-6">
```
deammer, line 300: |Before|After|
|---|---|
|Screen Shot 2021-07-21 at 11 46 53 AM|Screen Shot 2021-07-21 at 11 44 41 AM|

frontend/src/pages/offers/PalletsEditorSidebar.tsx
```diff
@@ -0,0 +1,121 @@
...
+  selectedLineItemId?: number
+  isLoading?: boolean
+  selectedPalletData?: PalletQuery
+  createPallet: () => void
+  selectPallet: (palletId: number) => void
+  addLineItem: (palletId: number) => void
+  selectLineItemId: (lineItemId: number) => void
+}
+
+const PalletsEditorSidebar: FunctionComponent<Props> = ({
```
deammer, line 21: I moved the entire sidebar from the PalletsEditor file to this one. No functional changes.

frontend/src/utils/data.ts
deammer: Logged in #267 for refactoring in the future.

PR #273, opened 2021-07-22
Delete line items when deleting a pallet

### LINKED ISSUE #272: deleting a pallet doesn't work if it has line items. We should show a confirmation dialog and then handle the deletions properly.

deammer:
Resolves #272.

Cascade the deletion of pallets to their associated line items. We should implement soft-deletion in the future, but not for v1.

### DIFF
--- a/src/models/line_item.ts
+++ b/src/models/line_item.ts
@@ -68,7 +68,10 @@ export default class LineItem extends Model<
   @Column
   public offerPalletId!: number
 
-  @BelongsTo(() => Pallet, 'offerPalletId')
+  @BelongsTo(() => Pallet, {
+    foreignKey: 'offerPalletId',
+    onDelete: 'CASCADE',
+  })
   public offerPallet!: Pallet
 
   @Column(DataType.STRING)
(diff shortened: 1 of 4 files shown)


src/models/line_item.ts
```diff
@@ -68,7 +68,10 @@ export default class LineItem extends Model<
   @Column
   public offerPalletId!: number
 
-  @BelongsTo(() => Pallet, 'offerPalletId')
+  @BelongsTo(() => Pallet, {
+    foreignKey: 'offerPalletId',
+    onDelete: 'CASCADE',
```
coderbyheart, line 73: @deammer this is all you need. It lets the database take care of deleting all entries that reference the pallet via a foreign key, when the pallet is deleted.

src/tests/pallets_api.test.ts
```diff
@@ -298,6 +312,8 @@ describe('Pallets API', () => {
       expect(res.errors).toBeUndefined()
       expect(await Pallet.findByPk(palletA.id)).toBeNull()
       expect(res.data?.destroyPallet?.pallets?.[0].id).toEqual(palletB.id)
+
+      expect(await LineItem.findByPk(lineItem.id)).toBeNull()
```
coderbyheart, line 316: **praise:** nice test!

coderbyheart (approved):

src/models/line_item.ts
deammer: That's amazing 🤯 thank you!

Issue #242, opened 2021-07-07
Shipments should accept a list of sending and receiving hubs


deammer:
Notes from a meeting with Dom and Elliot:
- we usually send from and to multiple hubs, so need a multi-select on the Shipment form
- Libby, Dom, and Elliot have added all our hubs to the staging database

In practical terms, we need to update the schema like this:
```graphql
# Before
type Shipment {
  sendingHubId: Int!
  sendingHub: Group!
  receivingHubId: Int!
  receivingHub: Group!
}

# After
type Shipment {
  sendingHubIds: [Int!]!
  sendingHubs: [Group!]!
  receivingHubIds: [Int!]!
  receivingHubs: [Group!]!
}
```

deammer:
I would love some help with the backend/database side of this. I can handle the frontend though.

coderbyheart:
Closed via #274

PR #274, opened 2021-07-22
Shipments should have many hubs

coderbyheart:
See #242


deammer (approved):
🚀

coderbyheart:
@deammer this is ready (I've consolidated the commits), however because of the current issues on GitHub we don't get Action runs right now. Waiting for that.

PR #287, opened 2021-08-08
Add guard against user id not being present

coderbyheart:
While testing with the frontend, I discovered (#288) that when I log-in the ID of my user was null:

```javascript
UserAccount {
  dataValues: { id: null, auth0Id: '' },
  _previousDataValues: { auth0Id: undefined },
  _changed: Set(1) { 'auth0Id' },
  _options: { isNewRecord: true, _schema: null, _schemaDelimiter: '' },
  isNewRecord: true
}
```

That lead to bugs (e.g. a new group that I had created did not have a captain).

This adds a check for this problem. Since it can't be solved by the user (or the application), `assert` is used.

### DIFF
--- a/src/resolvers/group.ts
+++ b/src/resolvers/group.ts
@@ -1,5 +1,6 @@
 import { Type } from '@sinclair/typebox'
 import { ApolloError, ForbiddenError, UserInputError } from 'apollo-server'
+import { strict as assert } from 'assert'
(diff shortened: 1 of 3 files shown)


coderbyheart:
Should use Assert instead of console

deammer (approved):

PR #300, opened 2021-08-15
Set up Clever Cloud deployment pipeline

deammer:
Documentation:
https://www.clever-cloud.com/doc/getting-started/quickstart/#deploy-your-code

### DIFF
--- a/db/config.json
+++ b/db/config.json
@@ -21,7 +21,7 @@
     "dialect": "postgres"
   },
   "production": {
-    "use_env_variable": "POSTGRESQL_ADDON_URI",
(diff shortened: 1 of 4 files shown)


coderbyheart:
I want to separate the deployment of the server and the static web application. There is no need to run the web app through express, because it is a static site.

coderbyheart:
I got a little further, however, when the server starts, it fails with 
```
ModelNotInitializedError: Model not initialized: Member "build" cannot be called. "UserAccount" needs to be added to a Sequelize instance.
```
which is weird, because in the build step no migration is detected.

coderbyheart:
Superseded by #314

PR #303, opened 2021-08-17
Deploy to sloppy.io

coderbyheart:
The server bundle can be configured via environment variables,
so we do not need to include any configuration


coderbyheart:
([Original post](https://distribute-aid.slack.com/archives/C0120QSNB2N/p1629323374034100?thread_ts=1629183540.033500&cid=C0120QSNB2N))

I've started to look into this, and didn't come far with my first attempt (compiling and providing a ZIP archive), however I then remember a thing called Docker :wink:
So what I am working on now is containerizing the backend and running it on a "Docker as a service"  thing called sloppy.io (they are a company from the Netherlands).
I got a DB image running, and the backend server only.

Now I am trying to initialize the database once through adding a second container, that only runs the npx sequelize-cli db:migrate --config ./db/config.json command.

image

coderbyheart:
I got it figured out, and I think a pretty flexible setup. I have a Docker image that runs once to apply migrations, one for the backend, one for the DB and finally I am packaging the frontend app in a container that is served by NGinx. Everything is running on Sloppy.io. I now need to add a new auth0 client (temporary to see get a log-in working).

Once that is done, sloppy.io supports updating the deployment via webhooks, so this would look like this (all orchestrated by GitHub Actions):
push to GitHub -> run tests -> make release -> update docker images -> publish updated docker images to docker hub -> trigger re-deploy of Sloppy

coderbyheart:
Got it to run with the current auth0 setup. Next up will be to trigger deployments from GitHub actions.

coderbyheart:
I'm not a fan of the dockerized solution, because it adds an additional layer of complexity, and I don't have enough experience with maintaining infrastructure using Docker.

PR #314, opened 2021-08-24
feat: switch over to passport.js

coderbyheart:
This adds username/password based authentication,
using a cookie when accessing secure routes

See [./docs/authentication.md](https://github.com/distributeaid/shipment-tracker/blob/passport.js/docs/authentication.md) for description of the implementation.

- [x] use email instead of username
- [x] add email field
- [x] add password reset
- [x] confirm email after registration
- [x] send emails
- [x] change frontend `useAuth` hook to use context, so auth state can be access everywhere
- [x] add `ShipmentSendingHub` / `ShipmentReceivingHub` to db schema again
- [x] figure out what broke sequelize
- [x] add CNAME to Clever Cloud cellar bucket so app works properly (e.g. index.html is served as DirectoryIndex)
- [ ] send email with password token on password reset
- [ ] new auth forms need some love
- [ ] auto-refresh token on user activity
- [ ] convert clever-cloud infrastructure to code, using the [Clever Cloud CLI](https://www.clever-cloud.com/doc/getting-started/cli/)
- [ ] add CLI for making users admins

CAPTCHA will be handled in a separate PR: #335 

Includes changes to host it on [CleverCloud](https://www.clever-cloud.com/en/), so this PR can be reviewed on a working instance.


ghost:
### Changes to your CodeSee Architecture Map:
View a CodeSee Map of these changes
#### Legend
CodeSee Map Legend

package.json
```diff
@@ -19,38 +19,44 @@
...
+    "nodemon": "2.0.12",
+    "passport": "0.4.1",
+    "passport-cookie": "1.0.9",
+    "pg": "8.7.1",
+    "prettier-plugin-organize-imports": "2.3.3",
+    "reflect-metadata": "0.1.13",
+    "sequelize": "6.6.2",
+    "sequelize-typescript": "2.1.0",
+    "supertest": "6.1.6",
+    "uuid": "8.3.2"
```
deammer: **praise:** thank you for pinning every dependency. I hope it wasn't too much of a pain.

src/authenticateRequest.ts
```diff
@@ -1,169 +1,91 @@
...
-
-export type RawAuthClaims = CommonAuthClaims & {
-  [ROLES_KEY]: string[]
+type AuthCookiePayload = {
+  /** user ID */
+  i: number
+  /** user is admin */
+  a: boolean
+  /** user hash */
+  c: string
```
deammer, line 12: **question (non-blocking):** are those fields named that way to match the way `passport-cookie` formats its data?

src/models/user_account.ts
```diff
@@ -25,8 +32,19 @@ export default class UserAccount extends Model<
 > {
   public id!: number
 
+  @Unique
   @Column
-  public auth0Id!: string
+  public username!: string
```
deammer: **question (non-blocking):** why require a username instead of an email?

src/routes/login.ts
```diff
@@ -0,0 +1,54 @@
...
+    if (user === null) {
+      // Penalize
+      await new Promise((resolve) => setTimeout(resolve, penaltySeconds * 1000))
+      return response.status(401).end()
+    }
+    if (!bcrypt.compareSync(valid.value.password, user.passwordHash)) {
+      // Penalize
+      await new Promise((resolve) => setTimeout(resolve, penaltySeconds * 1000))
+      return response.status(401).end()
+    }
```
deammer: **though (non-blocking):** this means that if I type the wrong username or password, I'll have to wait 10 seconds until I see an error message, right? I think that could become annoying from a UX perspective. I wonder if we could instead throttle access to that endpoint based on the user's IP (not necessarily in this PR though)

coderbyheart: Yes, a more advanced feature of this would be remembering the user, however, the point here is to slow down attackers, which most likely will switch IP addresses. I think a better solution will be to implement a CAPTCHA, like https://friendlycaptcha.com/.

src/models/user_account.ts
coderbyheart: Yes, good point. I started this switch without emails, but I realize we need them anyway (for password reset), so I will remove the username field and turn it into an email field.

src/authenticateRequest.ts
coderbyheart: No, but we encode this in the cookie sent to the browser, and there is a size limit, so I want to keep it as small as possible (while still using JSON).

package.json
coderbyheart: Only a little, not all of them are on the latest release (e.g. sequelize-typescript)

coderbyheart:
@deammer I realize that I've build the authentication API outside of GraphQL, as REST. Do you feel that is an issue? They are pretty independent of each other, any way (and were before).

src/routes/login.ts
deammer: I like the CAPTCHA solution 👍

deammer:
> @deammer I realize that I've build the authentication API outside of GraphQL, as REST. Do you feel that is an issue? They are pretty independent of each other, any way (and were before).

I feel like that's not a problem. The previous `/profile` endpoint was REST as well and we didn't have any issues.

coderbyheart:
I've asked friendlyCaptcha if they would provide us with a free business license, because only that one uses EU-only endpoints.

deammer (approved):

coderbyheart:
I have to figure out the problem with sequelize before merging.

PR #319, opened 2021-08-26
[WIP] add freebsd backend

dch:
tracking FreeBSD support

- needs some tweaking not to conflict with existing setups
- see `init_db` script and `start_postgresql` in `script/dev_setup`


coderbyheart:
Thanks @dch for offering, I think we have a working solution for now.

PR #335, opened 2021-09-09
Enable FriendlyCaptcha, sponsored by friendlycaptcha.com

coderbyheart:
- [ ] re-enable the FriendlyCaptcha implementation once #314 is in saga
- [ ] add badge to README and frontend as thank you for FriendlyCaptcha sponsoring the EU endpoint

From the CEO:

> Helping you with Friendly Captcha to protect your authentication methods to make donated aid from all over Europe possible seems to be the right thing to me.
> 
> That’s why we will make it possible for you!
> 
> Our EU Endpoint has some costs associated with it, which is why we are offering it to Business and Enterprise customers only. However, we would sponsor the EU Endpoint for you. A sponsor badge that tells the world that we are supporting you sounds good to us 🙂
> 
> We have just enabled the EU Endpoint for you 🙂


coderbyheart:
Implementation is quite far behind saga, so start from scratch. Copied info to a new issue: #459

PR #495, opened 2021-12-21
Address testing feedback from Session 1 and 2

coderbyheart:
This changes the implementation to fetch the
list of groups the user is captain of.
While the backend prevents the creation of more
than one group this implementation provides
some robustness against future changes.

The actual problem is that the backend
does not populate the groupId in the session
cookie, therefore it is always undefined.

Addresses more feedback from https://www.notion.so/distributeaid/Sara-and-Nicole-Testing-22-12-21-e6e28813c06f4dd9a0f54b47168af6ca


coderbyheart:
@deammer feel free to apply corrections if you see issues. No hard feelings!

PR #643, opened 2022-02-28
Pallet validation v2

coderbyheart:
Use volume and weight based validation.

See [the tests](https://github.com/distributeaid/shipment-tracker/pull/643/commits/1f384edca5be23067917ffdd0f931183cef1d342#diff-2a715405043f168e33e494abab359b1cf1fe64a19fc9fa1fbce1b0e3ce388206) for examples.

See 
- https://www.notion.so/distributeaid/It-might-be-worth-having-some-limits-or-flags-when-offers-are-over-a-certain-size-7ea49b440cf940b88ecbc37933aaaf50
- https://www.notion.so/distributeaid/1-pallet-can-contain-any-number-of-boxes-containers-depends-on-the-size-of-the-container-Is-1-36-23d67a3ab6464a09aa3aee100a727676

Screenshot_20220301_001252


frontend/src/pages/offers/LineItemForm.tsx
```diff
@@ -391,9 +392,7 @@ const LineItemForm: FunctionComponent<Props> = ({
           name="containerCount"
           type="number"
           min={1}
-          max={36}
```
deammer, line 394: **question:** do we not want to enforce a maximum amount anymore?

frontend/src/pages/offers/ViewPallet.tsx
```diff
@@ -90,11 +100,6 @@ const ViewPallet: FunctionComponent<Props> = ({
             </p>
             <ul className="list-disc list-inside">
               <li>a pallet can contain at most 1 bulk bag</li>
-              <li>a pallet can contain a maximum of 36 boxes</li>
-              <li>
-                if a pallet contains a bulk bag, it can also contain a maximum
-                of 18 boxes
-              </li>
```
deammer, line 97: **question:** same question as above. Do we not want to enforce those restrictions anymore?

deammer (approved):
I have a couple of non-blocking questions, but otherwise, this looks good to me, thank you for your time!

frontend/src/pages/offers/ViewPallet.tsx
coderbyheart: Yes, this limit was based on banana boxes, but other types of boxes must be possible. This was flagged in test feedback: https://www.notion.so/distributeaid/1-pallet-can-contain-any-number-of-boxes-containers-depends-on-the-size-of-the-container-Is-1-36-23d67a3ab6464a09aa3aee100a727676

Issue #675, opened 2022-03-17
Only make the pallet item editable that the user clicks "edit" on


RabeaGleissner:
Clicking the "edit" button on an item in a pallet currently shows all items in the form view.
Instead, only the current item should be editable.

![image](https://user-images.githubusercontent.com/8995723/158766823-f958f7f9-130e-4344-8114-84ec23a98909.png)

coderbyheart:
Closed via #715

PR #715, opened 2022-04-07
Only make the pallet line item editable that the user clicks "edit" on

RabeaGleissner:
Fix for https://github.com/distributeaid/shipment-tracker/issues/675

I also decided to add a test with React Testing Library which was a bit more complex than I expected due to the Apollo queries. Would be interested in opinions on that and if anyone can think of a better approach.

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -48,7 +48,7 @@
     "@babel/helper-environment-visitor": "7.16.7",
     "@babel/preset-env": "7.16.11",
     "@babel/preset-typescript": "7.16.7",
-    "@playwright/test": "1.20.0",
+    "@playwright/test": "1.20.2",
     "@testing-library/dom": "^8.11.3",
     "@testing-library/react": "13.0.0",
     "@testing-library/user-event": "^13.5.0",
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -43,12 +43,15 @@
     "yarn": ">= 1.22.5"
   },
   "devDependencies": {
+    "@apollo/react-testing": "^4.0.0",
     "@babel/core": "7.17.9",
     "@babel/helper-environment-visitor": "7.16.7",
     "@babel/preset-env": "7.16.11",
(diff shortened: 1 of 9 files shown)


frontend/src/pages/offers/LineItemForm.tsx
```diff
@@ -287,7 +287,7 @@ const LineItemForm: FunctionComponent<Props> = ({
   }
 
   return (
-    <form onSubmit={submitForm} className="pb-12">
+    <form onSubmit={submitForm} className="pb-12" aria-label="Add Line Item">
```
RabeaGleissner, line 290: React Testing Library prompted me to add the aria label because otherwise it couldn't find the form. Reading up on MDN it seems like an aria label or title should be added to a form.

https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/form_role

frontend/package.json
```diff
@@ -43,12 +43,15 @@
     "yarn": ">= 1.22.5"
   },
   "devDependencies": {
+    "@apollo/react-testing": "^4.0.0",
```
coderbyheart, line 46: **nitpick**: we use exact versions in this project, but this will be picked up by renovate.

frontend/src/pages/offers/LineItemForm.tsx
coderbyheart: **praise:** 👍🏻

frontend/tests/builders/index.ts
```diff
@@ -0,0 +1,5 @@
+import createLineItem from './createLineItem'
+import createOffer from './createOffer'
+import createPallet from './createPallet'
+
+export { createLineItem, createOffer, createPallet }
```
coderbyheart, line 5: **nitpick:** maintaining a barrel file is extra effort and does IMO no significant advantages. Consider removing it.

coderbyheart (approved):
Can merge as is.

PR #716, opened 2022-04-07
Set timezone for jest

RabeaGleissner:
The test in `FormattedDate.spec.tsx` was failing for me with the wrong time of day - probably because I'm in Singapore?

This is the failure I got:

Screenshot 2022-04-07 at 14 30 56

Setting the timezone before running jest fixes it!


coderbyheart:
Good catch!

